PR #8, opened 2017-02-02
5C-ZipLocation

chrisbrandow:
* set `IssuesContainerViewController` as `EditLocationViewControllerDelegate` & implemented methods
* dismiss VC on cancel
* set zipcode in UserDefaults on validated zip code entry & post notification
* set locationButton title
* add notification observation in `IssuesViewController` and post fetch with zipcode

Do we care about 9-digit zip codes?

### DIFF
--- a/FiveCalls/FiveCalls/IssuesViewController.swift
+++ b/FiveCalls/FiveCalls/IssuesViewController.swift
@@ -16,8 +16,8 @@ class IssuesViewController : UITableViewController {
         super.viewDidLoad()
         // edgesForExtendedLayout = []
         navigationController?.setNavigationBarHidden(true, animated: false)
-        
-        fetchIssues(forZip: nil)
+        let zip = UserDefaults.standard.string(forKey: UserDefaultsKeys.zipCode.rawValue)
+        fetchIssues(forZip: zip)
     }
     
     override var preferredStatusBarStyle: UIStatusBarStyle {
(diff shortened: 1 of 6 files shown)


subdigital:
Eventually we will probably need to support the +4 on zip codes, I figured this was easier to get out the door for now.  Also, most people will likely use the GPS to locate them anyway, so they don't end up with a split district response.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -17,7 +17,7 @@ class IssuesViewController : UITableViewController {
         // edgesForExtendedLayout = []
         navigationController?.setNavigationBarHidden(true, animated: false)
         
-        fetchIssues()
+        fetchIssues(forZip: nil)
```
subdigital: This will need to read from `UserDefaults`

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -26,14 +26,24 @@ class IssuesViewController : UITableViewController {
     
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        
-//        if let zip = UserDefaults.standard.string(forKey: UserDefaultsKeys.zipCode.rawValue) {
-//            
-//        }
+        NotificationCenter.default.addObserver(self, selector: #selector(zipCodeChanged(_:)), name: .zipCodeChanged, object: nil)
+        if let zip = UserDefaults.standard.string(forKey: UserDefaultsKeys.zipCode.rawValue) {
+            fetchIssues(forZip: zip)
```
subdigital: Ah I see you did it here. I don't think we need to reload issues every time the view appears, which gets called pretty frequently.  the data doesn't change so rapidly that we have to keep fresh data coming in as you switch back to the app, for instance.

subdigital:
👍

PR #13, opened 2017-02-03
Closes #4: Embed & Use Roboto Font

### LINKED ISSUE #4: Change app font to Roboto
In matching the style of the website, I'd like to embed the [Roboto](https://www.fontsquirrel.com/fonts/roboto) font and use it where appropriate.

Definitely for the copy in the walkthrough, but probably for all labels, buttons, view controller titles, etc. (we can probably accomplish most of this w/ appearance proxies).

Here's a link on [how to embed fonts on iOS](http://stackoverflow.com/questions/4969329/how-to-include-and-use-new-fonts-in-iphone-sdk).

bjtitus:
This embeds the Roboto Font and sets the Bold Condensed version as the global default. It takes into account font traits provided by the system so that Dynamic Text sizes are respected and font styles, like Italic, are utilized when possible.

I also changed the IssuesViewController to use dynamic cell sizing and all labels I could find to use the dynamic text styles so that they resize when the accessibility text size changes.

Here's a comparison of the text resizing:
napkin 02-02-17 10 53 55 pm

There are some drawbacks to this method:
- It is a little hidden and hard to understand without some serious experience with `UIAppearance` and `UIFontDescriptor`.
- Custom font selections will be overridden, unless `attributedText` is used


subdigital:
Nice!  I agree it can be a little weird if someone wants to change the font, but I think it's worth it to get dynamic text.

Issue #6, opened 2017-02-02
Save location by GPS


subdigital:
Tap Edit Location, tap "Use My Location".

- prompt to use location w/ `CLLocationManager`
- on callback, use lat/lng and reverse geocode to get address
- save these values to `UserDefaults`, preferrably under a single key w/ a dictionary value
- clear any saved zip code
- trigger a refresh (via notification or callback) when upon change
- `IssuesViewController` reads the saved location / address and passes to the fetch operation when loading issues.
- location button title should be changed to match the city from the saved location

subdigital:
One thing to keep in mind is that we'll have to clear out the saved zip when GPS is used, and clear out the saved GPS if zip is used.  The issues view controller will have to prefer using GPS over zip, but check both.

Alternatively we could use another key that specifies which location type was used:

```swift
let locationType = UserDefaults.standard.string(for: UserDefaultKeys.locationType.rawValue)
if locationType == "gps" {
  // use saved address
} else if locationType == "zip" {
  // use zip
} else {
  // nil location
}
```

Up to you.

chrisbrandow:
For now, I'm proceeding with "when in use" permissions only.

subdigital:
Yes, in fact I wish there was a “single use only” option :)

PR #14, opened 2017-02-03
Save location by GPS #6

chrisbrandow:
https://github.com/5calls/ios/issues/6

The primary logic that I have employed regarding whether to fetch with zip or latLong is that when a user sets zip or latLong, the other is deleted. So when a fetch is requested, the only non-nil value is passed along to the query. Additionally if both are nil, then the non-query url is passed along. This felt natural to do and less complicated than setting an order of preference or setting another key to indicate which to use. But definitely open to a different approach. I have confirmed that switching between location/zip as a user correctly returns the correct results based on last selected location/zip.

* added CLLocationManager to `EditLocationViewController`
* added "when in use" authorization to info.plist
* `IssuesContainerViewController`
    * lookup the returned location
    * refactored setTitle
    * refactored `updateWith` for zipCode and for lat/long
* updated `IssuesManager` to handle zip or latLong
* updated `FetchIssuesOperation to handle zip or latLong
    * updated execute to append zip or lat long -- confirmed both work

Lastly, there are a few dinky cleanup commits. I would recommend the "squash and merge" option, or I can squash locally and push it up fresh, but i added the commits after I noticed small things sequentially after making the PR.


subdigital:
Just a point of feedback on commit messages, I don't think it's valuable to reference the branch name as the subject of each commit.  Here's a good article on why:  http://chris.beams.io/posts/git-commit/

FiveCalls/FiveCalls/EditLocationViewController.swift
```diff
@@ -8,16 +8,25 @@
 
 import UIKit
 import CoreLocation
+//import GLKit
```
mccarron: Oh please tell us we're getting some cool 3D effects? ;)

FiveCalls/FiveCalls/EditLocationViewController.swift
```diff
@@ -29,14 +38,15 @@ class EditLocationViewController : UIViewController {
...
-        
+        locationManager.requestWhenInUseAuthorization()
     }
     
     @IBAction func cancelTapped(_ sender: Any) {
         delegate?.editLocationViewControllerDidCancel(self)
     }
     
     @IBAction func submitZipCodeTapped(_ sender: Any) {
+        //TODO: delete location if present
```
mccarron: Does this comment still make sense if we are deleting the defaults on updateWith below? Or did you mean in the controller itself?

FiveCalls/FiveCalls/IssuesContainerViewController.swift
```diff
@@ -65,13 +62,56 @@ class IssuesContainerViewController : UIViewController, EditLocationViewControll
     }
 
     func editLocationViewController(_ vc: EditLocationViewController, didSelectZipCode zip: String) {
+        dismiss(animated: true, completion: { //`updateWith` must be in completion block so that VC is listening for notification
+            self.updateWith(zipCode: zip)
```
mccarron: Should use [weak self] on completion blocks that reference self. (A few of them here below as well)

FiveCalls/FiveCalls/IssuesContainerViewController.swift
```diff
@@ -65,13 +62,56 @@ class IssuesContainerViewController : UIViewController, EditLocationViewControll
...
+    }
+
+    //Mark: private functions
 
+    private func setTitleLabel() {
+        if let zip = UserDefaults.standard.string(forKey: UserDefaultsKeys.zipCode.rawValue) {
+            locationButton.setTitle(zip, for: .normal)
+        } else if let locationInfo = UserDefaults.standard.value(forKey: UserDefaultsKeys.locationInfo.rawValue) as? [String: Any] {
+            let displayName = (locationInfo["displayName"] as? String) ?? "Selected Location"
+            self.locationButton.setTitle(displayName, for: .normal)
```
mccarron: In this case no need to use `self.` outside of completion blocks.

FiveCalls/FiveCalls/FetchIssuesOperation.swift
```diff
@@ -10,22 +10,35 @@ import Foundation
...
     
     lazy var sessionConfiguration = URLSessionConfiguration.default
     lazy var session: URLSession = { return URLSession(configuration: self.sessionConfiguration) }()
     
     override func execute() {
-        let url = URL(string: "https://5calls.org/issues/")!
+        var urlString = "https://5calls.org/issues/"
+        let urlQueryString: String? = zipCode ?? latLongString
+        if let query = urlQueryString {
+            urlString = urlString + "?address=" + query
```
bjtitus: `NSURLComponents` + `queryItems` might be a bit cleaner here.

FiveCalls/FiveCalls/IssuesContainerViewController.swift
chrisbrandow: Thanks. I actually hadn't fully considered the [weak self] issue. I've been making an unexamined assumption that Swift (3) was magically taking care of that for me.

FiveCalls/FiveCalls/FetchIssuesOperation.swift
chrisbrandow: Yeah, though perhaps it would be better to break that into a separate ticket though with a thorough clarification of the fetch operation. @subdigital, what do you think?

FiveCalls/FiveCalls/EditLocationViewController.swift
chrisbrandow: I'm sure we can find a reason that we(I) need to learn how to write shaders in order to complete this app.

FiveCalls/FiveCalls/IssuesContainerViewController.swift
```diff
@@ -65,13 +62,56 @@ class IssuesContainerViewController : UIViewController, EditLocationViewControll
...
+        locationInfo["latitude"] = location.coordinate.latitude
+        geocoder.reverseGeocodeLocation(location, completionHandler: { (placemarks, error) in
+            let prefix = placemarks?.first?.subThoroughfare ?? ""
+            let street = placemarks?.first?.thoroughfare ?? ""
+            let streetAddress = prefix + " " + street
+            locationInfo["displayName"] = streetAddress != " " ? streetAddress : nil
+            locationInfo["zipcode"] = placemarks?.first?.postalCode ?? ""
+            completion(locationInfo)
+        })
+    }
```
chrisbrandow: sounds great. that was my initial approach. no particular reason I didn't stick with it.

subdigital: ok let's get this ready for merge and we can make that change later.  Can you merge in master / resolve that conflict?

PR #26, opened 2017-02-05
Remove cell reload on viewWillAppear. Fixes #23

### LINKED ISSUE #23: Dynamic height rows seems to cause the issues list to shuffle
I noticed that once we supported dynamic height on the issues list, the table reshuffles strangely when scrolling.

Another (probably related) thing is this occasionally causes our labels to no longer wrap & cause their cell to increase in height, instead reverting to truncating with ellipses ('...').

bjtitus:
Removes what appears to be an unnecessary `reloadRows` call on `viewWillAppear`. @subdigital agreed that it seemed unnecessary. This call was causing rows to be reloaded when detail view controllers were popped and seemed to cause layout issues.

### DIFF
--- a/FiveCalls/FiveCalls/IssuesViewController.swift
+++ b/FiveCalls/FiveCalls/IssuesViewController.swift
@@ -29,6 +29,10 @@ class IssuesViewController : UITableViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         // NotificationCenter.default.addObserver(self, selector: #selector(locationChanged(_:)), name: .locationChanged, object: nil)
+    }
+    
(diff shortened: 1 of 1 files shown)


mccarron:
This call was added to be able to show that the Issues are refreshed after setting location as well as refresh if you had actually made the call with the checkmark.

bjtitus:
@mccarron I updated the PR to move this reload to `viewDidAppear`, which seems to keep the checkboxes working. I'm not sure I understand reloading the cells after setting a location.

Would it make sense to only reload `selectedIndexPath`? That solves the case of the checkmark being updated for completed issues.

mccarron:
We went with visible incase for whatever reason the issue text is different. But we didn't do a straight `reloadData()` because we wanted to preserve scroll position.

Issue #10, opened 2017-02-02
Share Issue on Twitter / Facebook / Email


subdigital:
- open share sheet, use URL to issue on 5calls.org
- TODO: Universal links?

PR #27, opened 2017-02-05
Added basic sharing support

alexanderjrobinson:
Added basic support for sharing issues on the issue detail and call script views for issue https://github.com/5calls/ios/issues/10.

I will create a separate issue for universal links because that will likely require a new operation to fetch a specific issue from the server based on id and I didn't want to hold up this feature.

### DIFF
--- a/FiveCalls/FiveCalls/IssueActivityItemSource.swift
+++ b/FiveCalls/FiveCalls/IssueActivityItemSource.swift
@@ -1,0 +1,45 @@
+//
+//  IssueActivityItemSource.swift
+//  FiveCalls
+//
+//  Created by Alex on 2/4/17.
+//  Copyright © 2017 5calls. All rights reserved.
+//
+
+import UIKit
+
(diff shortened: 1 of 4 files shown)


FiveCalls/FiveCalls/IssueActivityItemSource.swift
```diff
@@ -0,0 +1,45 @@
...
+    init(issue: Issue) {
+        self.issue = issue
+    }
+    
+    func activityViewControllerPlaceholderItem(_ activityViewController: UIActivityViewController) -> Any {
+        return issue.name
+    }
+    
+    func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivityType) -> Any? {
+        return "I just called my rep to \(issue.name) — you should too: http://5calls.org/#issue/\(issue.id)?utm_campaign=twshare"
```
subdigital: @nickoneill just a heads up regarding the share message text.  Do you think this grammar assumption is okay to make?

PR #39, opened 2017-02-07
Call Script Controller Revisions

mccarron:
* Allows you to report a call response even if you reported it before
* Navigation now pops back one level once all people are contacted vs all the way to root
* Logs phone number to the Report API including alternate phone numbers when used
* Alternate phone number popup will dismiss with any touches outside the view including scrolling.
* Fixed some layout issues on a smaller iPhone screen

### DIFF
--- a/FiveCalls/FiveCalls/CallScriptViewController.swift
+++ b/FiveCalls/FiveCalls/CallScriptViewController.swift
@@ -49,13 +49,8 @@ class CallScriptViewController : UIViewController, IssueShareable {
 
     func callNumber(_ number: String) {
         lastPhoneDialed = number
-        if let dialURL = URL(string: "telprompt:\(number)") {
-            if #available(iOS 10.0, *) {
-                UIApplication.shared.open(dialURL)
-            } else {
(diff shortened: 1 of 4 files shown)


FiveCalls/FiveCalls/CallScriptViewController.swift
```diff
@@ -35,48 +31,27 @@ class CallScriptViewController : UIViewController, IssueShareable {
...
-        print("dialing \(number)")
+        lastPhoneDialed = number
         if let dialURL = URL(string: "telprompt:\(number)") {
             if #available(iOS 10.0, *) {
-                UIApplication.shared.open(dialURL) { success in
-                    //Log the result
-                }
+                UIApplication.shared.open(dialURL)
             } else {
                 UIApplication.shared.openURL(dialURL)
```
subdigital: We have this check in two places. Mind moving to a category on UIApplication, something like:

```swift
UIApplication.shared.fvc_open(_ url:...)
```

?

mccarron: Can do 👍

Issue #42, opened 2017-02-07
About Screen: Social Links don't unselect


mccarron:
After selecting the various links in the Social section of the About screen they don't get un-highlighted.

![img_9117](https://cloud.githubusercontent.com/assets/45654/22700760/4354180e-ed21-11e6-99c2-ede31c736eb3.PNG)

PR #43, opened 2017-02-07
#42 - About Screen: Social Links don't unselect

chrisbrandow:
Deselects the selected tableViewCell after performing designated action.


subdigital:
Curious why this isn't working by default...

https://developer.apple.com/reference/uikit/uitableviewcontroller/1614758-clearsselectiononviewwillappear

mccarron:
Guessing viewWill/DidAppear is not getting called? It works for the other
sections which transition away to a new VC.  Those Social ones open a
system screen.

subdigital:
Good call. I’ll bite.

PR #51, opened 2017-02-09
Analytics Calls Hooked In

mccarron:
Added a bunch of custom actions to log the user behavior across the app.


subdigital:
Looks good, thanks

Issue #59, opened 2017-02-10
Flashing Images in `IssueDetailViewController`


chrisbrandow:
every time the view appears, the `avatarImage` flashes.

mccarron:
Forgot to note this also happens in `CallScriptViewController` and should also be fixed along with this.

subdigital:
Resolved by #60

PR #60, opened 2017-02-10
#59 Fixed the image flashing in the `IssueDetailViewController` avatars

chrisbrandow:
https://github.com/5calls/ios/issues/59

This solution is memory only, and is local to this `IssueDetailViewController`, but could be moved to a singleton/app delegate.
Or, conversely, we could use an existing cache solution. However, this seems like a small thing to pull in another dependency for.

*  `[String: UIImage]` property to `IssueDetailViewController`
    * check it when setting `cell.avatarImageView.image`
    * or set it after downloading remote image
* added `setRemoteImage` method with a "preferredImage" argument

### DIFF
--- a/FiveCalls/FiveCalls/IssueDetailViewController.swift
+++ b/FiveCalls/FiveCalls/IssueDetailViewController.swift
@@ -15,7 +15,8 @@ class IssueDetailViewController : UIViewController, IssueShareable {
     var issuesManager: IssuesManager!
     var issue: Issue!
     var logs: ContactLogs?
-    
+    internal var images = [String: UIImage]()
+
     @IBOutlet weak var tableView: UITableView!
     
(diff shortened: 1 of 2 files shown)


mccarron:
This solves the issue on `IssueDetailViewController` but we still have the issue with the avatar flashing on the `CallScriptDetailViewController`. 

So maybe we should instead of AppDelegate or a Singleton just use `NSCache` and have `RemoteImageView` handle the caching completely taking that logic out of the view controllers at all?

subdigital:
Definitely would prefer using `NSCache` or similar for this sort of thing because in low memory conditions it will automatically get flushed, whereas we'd have to handle that case here. I'll update this PR w/ some modifications.

subdigital:
@chrisbrandow thanks for starting this off!

Issue #20, opened 2017-02-04
Extract strings in preparation for localization


subdigital:
Currently the app uses hard-coded strings.  We're prepping to localize everything into some other languages.  The first step is to extract the strings in the app into strings files so we can hand these off to translators.

alexanderjrobinson:
What do you think about using something like R.swift for the localization strings?

subdigital:
Not familiar with it, looking now.

subdigital:
Tracking this in #61

PR #61, opened 2017-02-10
Extract strings for localization. Addresses #20

erikkrietsch:


mccarron:
I read "tomfoolery" and knew it was `misplaced="YES"`

mccarron:
Looks good to me but I'll let @subdigital take a look first.

subdigital:
@iosjillian heads up on this PR before you start on any localization effort

subdigital:
@erikkrietsch @iosjillian

Pardon my ignorance here, but I don't have a lot of experience with localization. I have some workflow questions about this:

### Adding New strings

- does this have to be done manually?  For instance, I add a new user facing message in the app, I manually add the key to `Localizable.strings` for the en-US locale, and then add the key to the enum to reference this?  What happens to the comment specified in the `.localized` helper function?

### Submitting for translation

- We have 4 localizable strings files, 3 for storyboards and 1 for the code in our app. What's the workflow like for submitting all of these for translation? 
- What's the process like for importing completed translations?

erikkrietsch:
No apologies necessary. I really only have a little experience myself. In fact, I would like to tap @leopic to fill in some of the gaps in my knowledge. 

To answer your questions, when adding new user-facing strings via code, then yes, you have to add that string manually to the `Localizable.strings` file, then add the key and enum in the appropriate places. I decided to go the route of enums because I generally can't stand seeing string literals, and like to remove as many possibilities for typos as possible. There are some different schools of thought about how to approach this, and if you'd like to change the interface, I can accommodate. An example of a different approach would be something like:
```Swift 
"This is a string to be translated.".localize()
``` 
and the key in the `Localizable.strings` file would be the string itself, and the value would be identical. Like this: 
```
"This is the string to be translated" = "This is the string to be translated";
```
This makes it perhaps a little easier to understand at the call sight, but it leaves all those string literals I dislike so much. But I don't want to push my personal preferences onto the project, so I'll leave it to you to decide.

The `comment` argument is something that can be added to help aid in the translation of the string. Because languages can be idiomatic and sometimes vague, especially to non-native speakers, adding a comment can help for an ambiguous turn of phrase. Some people like to add comments to all the strings that are being translated, while I tend to prefer them only when the phrase is ambiguous or otherwise difficult to understand. 

As far as submitting for translation, there's a fairly direct explanation here: https://littlebitesofcocoa.com/183-importing-exporting-xliff-translation-files This is where I should to mention I've never actually done that part, and would like to ask @leopic to chime in, if he has any additional comments.

subdigital:
As for the comment argument, I understand what it's use is, but I don't see how that makes it into a file that translators would use.  Is there an automated tool that extracts this for us?

erikkrietsch:
Off the cuff, my guess is it gets included in the XLIFF file that gets generated when you export for localization from Xcode. As far as a tool--like a 3rd party CLI--that can do this for us, I'm not sure. Maybe [this is useful](https://github.com/vmalyi/fastlane-plugin-localization). Xcode can do it, and I'm pretty sure you only have to export/import once per language translation.

iosjillian:
@subdigital @erikkrietsch 

Re: String literals, [R.swift](https://github.com/mac-cain13/R.swift) can replace all of those. I agree, I'm not a fan of the literals either. R.swift will basically do the same thing as your enum, but will be automatically generated based on the strings files. Code will look like this:

```
// Localized strings are grouped per table (.strings file)
let welcomeMessage = R.string.localizable.welcomeMessage()
let settingsTitle = R.string.settings.title()

// Functions with parameters are generated for format strings
let welcomeName = R.string.localizable.welcomeWithName("Alice")
```

R.swift was added to the project in PR #65

iosjillian:
@subdigital @erikkrietsch 

Since iOS and Android (and presumably web) will share string translations, it may make sense to manually keep all the desired strings in a spreadsheet instead of worrying about generating that info from what's in the iOS code. We may want to match the apps to the strings source, not the other way around.

designatednerd:
FWIW [Twine](https://github.com/mobiata/twine) can use a single text file for translations and dump out iOS `Localizable.strings`, Android `strings.xml` and a couple other formats that could work for the web.

subdigital:
To get this one merged in, we'll have to merge in `master`. Also I'd like to move the string enum over to R.swift.

Workflow questions don't have to hold this PR up.

erikkrietsch:
Rebased on top of master.

erikkrietsch:
Rebased on top of master. I haven't moved the enum into R.swift yet. I'll get acquainted with the R API and take care of it.

erikkrietsch:
Refactored to use R.swift and squashed everything down to one commit.

PR #65, opened 2017-02-12
R.swift

iosjillian:
## Description 

- Added [R.swift](https://github.com/mac-cain13/R.swift) to the project (this process includes a run script). Note that when you add new resources you'll need to run the project to generate the references, not just build.
- Updated all "stringly typed" resources except for Strings, since those are all localized in PR #61.
- Updated the ReadMe with a blurb about R.swift and a reminder to run `bundle install` when getting started. (I'll be happy to update with anything else y'all think should be added!)

Let me know if you have any questions! This isn't necessary, but will help with localization, adds extra type safety, and it's just nice, y'know?


subdigital:
> Note that when you add new resources you'll need to run the project to generate the references, not just build.

@iosjillian Are you sure that's true? The Run Script Build Phase should be done before we compile sources right?

subdigital:
(also, why the hell haven't I used R.swift before, this is awesome)

iosjillian:
@subdigital I agree that's how it should work, but in practice it seems to be a little finicky. I just tested it out, and looks like building twice works if building once doesn't detect the change ¯\_(ツ)_/¯

Issue #47, opened 2017-02-08
Show Checkmark On Call Script Screen 


mccarron:
There is no indicator on that screen to show if you've contacted that contact or not.

One suggestion is a `CheckboxView` in a titleView along with `Contact X of Y`.

bengottlieb:
would we want to have some sort of on-device storage in that case, so we can keep track of these calls?

mccarron:
@bengottlieb We do already, we're storing call data with Pantry using the class ContactLogs to log when they tap a call outcome by issue ID and contact ID. ContactLog has a helper `hasContacted`

You can see it used on the main screen: 
https://github.com/5calls/ios/blob/master/FiveCalls/FiveCalls/IssuesViewController.swift#L109

bengottlieb:
Cool, didn't realize that's what Pantry was. Added PR #69 to address this, but uncovered another issue

PR #69, opened 2017-02-13
Addresses #47 "Show Checkmark On Call Script Screen" 

bengottlieb:
This is for issue #47 

I added a new method to ContactLogs to fetch the actual method of contact recorded for a given issue/contact combination, and refactored hasContacted() to use it.

I added colors for the selected state for BlueButtons, and then set the selected state when entering the CallScript controller, based on the recorded method of contact.


subdigital:
Thanks, @bengottlieb!

PR #71, opened 2017-02-14
updated unit tests for latest issues list

bengottlieb:
This should fix the problem in build #79


mccarron:
Haha we gotta find a better way to do that.

bengottlieb:
maybe access the issues list directly in the test script before looking for UI text?

PR #72, opened 2017-02-14
Seeded data for testing

subdigital:
- Adds a level of indirection for getting a `URLSession` by funneling through `URLSessionProvider.
- Checks for `UI_TESTING` process arguments, returning a different URLSession when it is found
- `SeededURLSession` builds `SeededDataTask` that look for special keys (like `GET:/issues`) in order to load up fixture data and call the completion handler with the data from this file.

Some notes:

- this only works on the simulator right now because it relies on paths part of the test bundle being accessible at runtime in the app. This might be avoided if we loaded the contents of the file into the `launchArguments` dictionary, but this feels like abusing the feature, and has gotchas of its own (like encoding certain characters like `=`). Open to feedback on this

- We probably want to distinguish from multiple calls to the same resource in order to provide different data, perhaps by matching on querystring. The HTTP method already had to be pushed into the key so we could have `GET` and `POST` on the same `/report` endpoint, but I think perhaps a metadata file (plist maybe) could provide all the info we need to match on, as well as response things like status code & headers.

For now though, UI Tests run w/o hitting the network and come back with consistent data.


subdigital:
@bengottlieb thoughts on this?

bengottlieb:
Very slick, I like it!

nickoneill:
👍 I've never seen this level of test abstraction for iOS before but it looks great!

PR #73, opened 2017-02-14
Code consolidation, human-readable colors

designatednerd:
- Added a couple extensions for things that are happening more than once
- Consolidated the colors so it's easier to a) see what they are and b) easier to read them in the code. 

Here's a screenshot of the named colors in a playground:

screen shot 2017-02-13 at 10 29 17 pm

There's some ever so slight shade differences between the two light blues and two dark blues, so I wasn't sure if that was purposeful or not. If I should consolidate those, let me know. 

Note: I prefixed methods in extension classes with `fvc` for two reasons: 
- Due to [a few really stupid issues with private method naming and extensions on Objective-C classes](https://pspdfkit.com/blog/2016/surprises-with-swift-extensions/)
- I have found in my own code it makes it a hell of a lot easier to distinguish during code review when I'm using methods added in a given project vs. what comes from AAPL.

### DIFF
--- a/FiveCalls/FiveCalls/AboutViewController.swift
+++ b/FiveCalls/FiveCalls/AboutViewController.swift
@@ -107,8 +107,7 @@ class AboutViewController : UITableViewController, MFMailComposeViewControllerDe
     }
 
     func showWelcome() {
-        let welcomeStoryboard = R.storyboard.welcome()
-        let welcomeVC = welcomeStoryboard.instantiateInitialViewController()! as! WelcomeViewController
+        let welcomeVC = R.storyboard.welcome.welcomeViewController()!
         welcomeVC.completionBlock = { [weak self] in
             self?.dismiss(animated: true, completion: nil)
         }
--- a/FiveCalls/FiveCalls/Extensions/UIColor+FiveCalls.swift
+++ b/FiveCalls/FiveCalls/Extensions/UIColor+FiveCalls.swift
@@ -37,9 +37,7 @@ extension UIColor {
     //MARK - Grays
     
     convenience init(fvc_gray component: Float) {
-        self.init(colorLiteralRed: component,
-                  green: component,
-                  blue: component,
+        self.init(white: CGFloat(component),
                   alpha: 1.00)
     }
     
@@ -48,6 +46,8 @@ extension UIColor {
     static let fvc_lightGray = UIColor(fvc_gray: 0.90)
     
     static let fvc_mediumGray = UIColor(fvc_gray:0.88)
+    
+    static let fvc_darkGray = UIColor(fvc_gray: 0.4)
     
     //MARK: - Other colors
     
--- a/FiveCalls/FiveCalls/Extensions/UIColor+FiveCalls.swift
+++ b/FiveCalls/FiveCalls/Extensions/UIColor+FiveCalls.swift
@@ -1,0 +1,63 @@
+//
+//  UIColor+FiveCalls.swift
+//  FiveCalls
+//
+//  Created by Ellen Shapiro on 2/13/17.
+//  Copyright © 2017 5calls. All rights reserved.
+//
+
+import UIKit
+
+// Copy and paste into a playground to be able to see all colors. 
+extension UIColor {
+
+    //MARK: - Blues
+    
+    static let fvc_lightBlue = UIColor(red:0.68,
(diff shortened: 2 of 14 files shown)


FiveCalls/Pods/DropDown/DropDown/helpers/DPDKeyboardListener.swift
```diff
@@ -36,12 +36,12 @@ extension KeyboardListener {
 		NotificationCenter.default.addObserver(
 			self,
 			selector: #selector(keyboardWillShow(_:)),
-			name: NSNotification.Name.UIKeyboardWillShow,
```
mccarron: Should we maybe avoid changing stuff in  `/Pods` since we're modifying a third party library and this will just get overwritten on an update?

FiveCalls/FiveCalls/AboutViewController.swift
```diff
@@ -107,8 +107,7 @@ class AboutViewController : UITableViewController, MFMailComposeViewControllerDe
     }
 
     func showWelcome() {
-        let welcomeStoryboard = R.storyboard.welcome()
-        let welcomeVC = welcomeStoryboard.instantiateInitialViewController()! as! WelcomeViewController
+        let welcomeVC = R.storyboard.welcome.welcomeViewController()!
```
subdigital: learning new things all the time :)

FiveCalls/FiveCalls/Extensions/UIColor+FiveCalls.swift
```diff
@@ -0,0 +1,63 @@
...
+    
+    
+    //MARK - Grays
+    
+    convenience init(fvc_gray component: Float) {
+        self.init(colorLiteralRed: component,
+                  green: component,
+                  blue: component,
+                  alpha: 1.00)
+    }
```
subdigital: This could be removed and just use `UIColor(white: xxx, alpha: 1.0)`

mccarron:
Been wanting to do this for a while, thanks for doing it. Will be easier going forward. 

I think consolidating those colors would be fine, I was typically grabbing them off with an eye dropper from the web and bet they grabbed slightly different colors.

I'd say revert the changes to the Pod and see what @subdigital has to say.

FiveCalls/Pods/DropDown/DropDown/helpers/DPDKeyboardListener.swift
```diff
@@ -36,12 +36,12 @@ extension KeyboardListener {
...
+			name: .UIKeyboardWillShow,
```
subdigital: You're editing a pod here, I would rather contribute this change upstream.

FiveCalls/Pods/DropDown/DropDown/helpers/DPDKeyboardListener.swift
subdigital: Agree, that should be contributed upstream, if we care to.

subdigital:
The colors were taken directly from the web, so I think they are fine for now.  The benefit is that they are way easier to change now, so 🎉

FiveCalls/FiveCalls/AboutViewController.swift
mccarron: Right? Really liking R.

designatednerd:
Whoops, missed that a couple of the changes were to pods. Stupid scope bar. Will fix in the morning.

FiveCalls/FiveCalls/Extensions/UIColor+FiveCalls.swift
```diff
@@ -37,9 +37,7 @@ extension UIColor {
     //MARK - Grays
     
     convenience init(fvc_gray component: Float) {
-        self.init(colorLiteralRed: component,
-                  green: component,
-                  blue: component,
+        self.init(white: CGFloat(component),
                   alpha: 1.00)
```
subdigital: Why not just get rid of this convenience init altogether and inline the `white:` calls?  We'll get the CGFloat conversion for free there anyway.

subdigital:
Rebased and made that change, this is now in `master`.  Thanks, @designatednerd!

designatednerd:
Sorry, got a little swamped at work. Thanks for handling @subdigital!

PR #74, opened 2017-02-14
Fastlane beta improvments

subdigital:
- automatically increments build number for beta builds and commits this automatically
- badges the app icon with "BETA" and the version number. These changes are reverted afterward.
- tags the version & pushes tags on completion
- pauses while you edit the changelog
- external distribution by default

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -163,12 +163,6 @@ GEM
       jwt (~> 1.5)
       multi_json (~> 1.10)
     slack-notifier (1.5.1)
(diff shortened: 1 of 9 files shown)


Gemfile.lock
```diff
@@ -11,6 +11,10 @@ GEM
     addressable (2.5.0)
       public_suffix (~> 2.0, >= 2.0.2)
     babosa (1.0.2)
+    badge (0.7.1)
```
mccarron: Totally trying that for my Betas.

Issue #40, opened 2017-02-07
Local weekday reminder to make your 5 calls


subdigital:
- toggle from welcome screen and settings menu
- schedules local notification for weekdays at a specified time
- on launch schedule the next one if needed

chrisbrandow:
Not entirely clear where it is launching from. I don't see the settings menu. Assuming it's not there yet.

subdigital:
There’s an “About” screen in the footer, it can go in there.  Maybe we need to rename the button, open to suggestions. (Right now “Settings" feels inappropriate because it’s all informational)

bengottlieb:
Is this something that needs to be added? I don't see any code in there for local notifications. Should we add the "feature" tag?

subdigital:
Yes, but I believe @chrisbrandow is working on this already.

chrisbrandow:
sorry for delay. got held up with a few things.

PR #77, opened 2017-02-14
#40 - Local weekday reminder to make your 5 calls

chrisbrandow:
Implemented to allow user to selectively choose weekdays
* created a multipleSelection view
* added a time selector, but the same time is applied to all


chrisbrandow:
simulator screen shot feb 14 2017 3 59 04 pm

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -34,7 +34,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if !UserDefaults.standard.bool(forKey: UserDefaultsKeys.hasShownWelcomeScreen.rawValue) {
             showWelcome()
         }
-        
+
+        var types: UIUserNotificationType = UIUserNotificationType()
```
subdigital: can remove the type specifier here: `var types = UIUserNotificationType()`

FiveCalls/FiveCalls/MultipleSelectionSegmentedControl.swift
```diff
@@ -0,0 +1,90 @@
...
+    func setSelectedButtons(at indices: [Int]) {
+        for index in indices {
+            buttons?[index].isSelected = true
+            let titleColor = (buttons?[index].isSelected)! ? deSelectedColor : selectedColor
+            buttons?[index].setTitleColor(titleColor, for: .normal)
+            buttons?[index].backgroundColor = (buttons?[index].isSelected)! ? selectedColor : deSelectedColor
+        }
+    }
+
+    override func awakeFromNib() {
```
subdigital: good habit to keep `super` in here, even if we know it's a no-op

FiveCalls/FiveCalls/MultipleSelectionSegmentedControl.swift
```diff
@@ -0,0 +1,90 @@
...
+            stackView.leftAnchor.constraint(equalTo: leftAnchor, constant: 0.0),
+            stackView.rightAnchor.constraint(equalTo: rightAnchor, constant: 0.0),
+            stackView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: 0.0),
+            ])
+        buttons?.forEach({ color(button: $0) })
+    }
+
+    @objc func buttonAction(_ sender: UIButton) {
+        sender.isSelected = !sender.isSelected
+        delegate?.control(changed: self)
```
subdigital: I don't think we need a delegate here. We can leverage `UIControl`'s `sendAction:forEvent` and just use `.valueChanged`. Consumers would use it just like `UISegmentedControl`.

FiveCalls/FiveCalls/ScheduleRemindersController.swift
```diff
@@ -0,0 +1,80 @@
...
+    override func viewDidLoad() {
+        daysOfWeekSelector.delegate = self
+        notificationsChanged = false
+        if let notifications = UIApplication.shared.scheduledLocalNotifications {
+            daysOfWeekSelector.setSelectedButtons(at: indices(from: notifications))
+            if let date = notifications.first?.fireDate {
+                timePicker.setDate(date, animated: true)
+            }
+        }
+        timePicker.setValue(UIColor(red:0.12, green:0.47, blue:0.81, alpha:1.00), forKey: "textColor")
```
subdigital: heads up there is a PR to move all these colors into static properties, so we'll have to change this once that is merged in

FiveCalls/FiveCalls/MultipleSelectionSegmentedControl.swift
```diff
@@ -0,0 +1,90 @@
...
+import Foundation
+
+protocol MultipleSelectionDelegate {
+    func control(changed control: MultipleSelectionControl)
+}
+class MultipleSelectionControl: UIControl {
+
+    //TODO: get values from persistence
+    //TODO: persist values
+    //TODO: override UIControl
```
subdigital: Seems like persistence is being handled by the local notification API?  do we need anything else here?

FiveCalls/FiveCalls/ScheduleRemindersController.swift
```diff
@@ -0,0 +1,79 @@
...
+//
+
+import Foundation
+
+class ScheduleRemindersController: UIViewController {
+
+    @IBOutlet weak var timePicker: UIDatePicker!
+    @IBOutlet weak var daysOfWeekSelector: MultipleSelectionControl!
+
+    var notificationsChanged: Bool?
```
iainsmith: Nit: This immediately get's set to false in viewDidLoad, so it could probably be
```swift
var notificationsChanged: Bool = false
```

this also simplifies the guard in viewWillDisappear to

```swift
guard notificationsChanged == true else { return }
```

chrisbrandow: Good catch! <smacks forehead>

mccarron:
Tagging issue #40

chrisbrandow:
I'm happy to resolve the conflict, but I wasn't sure if I should be the one to do that or not. I figured whoever merged this would "want" that task.

subdigital:
Nope, this all happens on the branch, so tag you’re it :)

subdigital:
And since you’re the only one w/ this branch you can rebase on master to essentially _pretend_ that you just made the branch. You’ll often get conflicts for the things you’ve changed that master also changed, but those are usually easy to resolve.

Once done you have to `git push -f` (or more safely `git push --force-with-lease`, which I literally just learned about today)

chrisbrandow:
merge conflicts fixed.

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -30,7 +30,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if !UserDefaults.standard.bool(forKey: UserDefaultsKeys.hasShownWelcomeScreen.rawValue) {
             showWelcome()
         }
-        
+
+        var types = UIUserNotificationType()
+        types.insert(.alert)
+        types.insert(.badge)
+        UIApplication.shared.registerUserNotificationSettings(UIUserNotificationSettings(types: types, categories: nil))
```
subdigital: `UIUserNotification*` is an iOS 10-only framework. We'll have to include back-ported support for iOS 9 devices.

According to this chart:

image

We have roughly3% of our devices that run iOS 9. (these numbers are just those that opt-in to iTunes Analytics)

Maybe a case could be made to make this iOS 10 only, but for now I think it's easy enough to keep it iOS 9 compatible.

subdigital: Also, I'd like to move the prompt until the user knows what it is for.

Here's my thought:

- notifications are off by default
- on the welcome screen, just above the 'get started' button, there's a UISwitch that says "Remind me to call", when toggled on, the prompt is displayed. This also pushes the schedule VC.
- if they skip this and just press Get Started, we have another opportunity to present it...
- on the menu -> reminder vc, add a UISwitch in the right nav bar area. When _off_ gray out the entire view (ooh maybe a blur overlay would look nice here).  When _on_, we can prompt for notifications then, and remove the blur to reveal the schedule settings.

(If we don't like the switch on the welcome screen, then maybe a tooltip when they first launch that mentions they can be reminded via the menu...)

Thoughts on this?

(/cc: @mccarron )

chrisbrandow: 1. Agreed about needing ios9 support. that was sloppy.
2. Agreed, notifications should be off by default and not requested until user acts as you suggest. I forget when building these system permissions thing because it only asks me once.
3. I like the idea of presenting it to them early, but an alternative might also be to wait and present a toolTip on the **2nd** time they have used the app. It might be less likely to bother lookie-loo users concerned about privacy, and more likely to be accepted by users demonstrating a minimal commitment to using the app? I don't think that is an overly complex logic.
4. I assume that this we would still keep the entry in the `aboutViewController`.

Perhaps the best way forward would be to move the `UIApplication.shared.registerUserNotificationSettings()` call to the `ScheduleRemindersController` and trigger it on first user action in that screen.

then merge that, and put the rest of these items into a new "issue". I would be fine with someone else doing those as well. But whatever you guys think best.

@mccarron , @subdigital

chrisbrandow: actually, @subdigital isn't [UIUserNotification](https://developer.apple.com/reference/uikit/uiusernotificationaction) the framework that was *deprecated* in iOS 10? I was trying to avoid the new [UNNotification](https://developer.apple.com/reference/usernotifications/unnotification) framework.

subdigital:
Ahhh, these are too similar. Yes, avoid UNUserNotification unless on iOS 10. Derp.

FiveCalls/FiveCalls/AppDelegate.swift
designatednerd: `UserNotifications` is the new framework, the old stuff was actually part of UIKit. 

I have some [sample code supporting both 9 and 10](https://github.com/designatednerd/iOS10NotificationSample/) you can yoink if you want.

subdigital:
@chrisbrandow are you still working on this?  I think the tooltip option for reminding users on 2nd launch is the best path forward.  I'd also like to have the switch at the right nav bar item.

If you don't have the time lmk, I can take this over.

PR #80, opened 2017-02-16
Added support for the iPad

bengottlieb:


FiveCalls/FiveCalls/IssueDetailViewController.swift
```diff
@@ -44,8 +44,12 @@ class IssueDetailViewController : UIViewController, IssueShareable {
             tableView.deselectRow(at: indexPath, animated: true)
         }
         logs = ContactLogs.load()
-        
-        navigationController?.setNavigationBarHidden(false, animated: true)
+
+        if let nav = self.navigationController, nav.viewControllers.count > 1 {
+            navigationController?.setNavigationBarHidden(false, animated: true)
+        } else {
+            tableView.contentInset = UIEdgeInsets(top: 64, left: 0, bottom: 0, right: 0)
```
designatednerd: any idea why the table view isn't picking up the content inset properly from the nav bar? In theory, you shouldn't have to hard-code this.

bengottlieb:
'cause that's not a navbar above it; it's the shared header from the Issues List page. I tried to change as little as possible, so I'm still using a lot of the same architecture from the iPhone version.

FiveCalls/FiveCalls/IssueDetailViewController.swift
bengottlieb: 'cause that's not a navbar above it; it's the shared header from the Issues List page. I tried to change as little as possible, so I'm still using a lot of the same architecture from the iPhone version.

designatednerd: Ah, ok, missed that. Might be worth seeing if you can grab the height from the view rather than hard-coding it, but `¯\_(ツ)_/¯`

bengottlieb: probably, but I think that would just trade one dependency (a hard coded number) with another (a way to get that value up-and-down the view controller hierarchy). I'm still not sure this design is the approach we want to take for the iPad version. or do we just keep the header in the left-hand controller.

Issue #78, opened 2017-02-15
Split Congressional District alert with Use My Location option


jablair:
I'm getting the Split Congressional District alert even when I use the Core Location option in lieu of entering my Zip code. If there's something about my location that makes the lookup unreliable, perhaps the dialog should be reworded?

I can see about collecting some additional info later for you folks.

bengottlieb:
Can you provide your lat/long? That might help track it down.

jablair:
<+38.88163057,-77.10497088>

If I take the lat/long down to <= 3 digits of accuracy to the right of the decimal point, I get a non-split ticket back from the API and it looks like the correct info.

bengottlieb:
This looks like a Google Civic Information problem; putting those values into the API directly gives me different data (and less) than entering in truncated values. I think we'll get enough accuracy anywhere in the country using just 3 significant digits (I tried it in Manhattan, which I would imagine is the densest in the country). I also tried a few other lat/longs, and while I usually got my reps, I sometimes got more (state level) data when I used fewer digits.

subdigital:
Chopping off significant digits changes the accuracy of the location, so I think this would yield inconsistent results. This is going to have a much different effect for someone in Texas than in Maine.

@jablair can you try entering in your address to the API directly?

```
http://5calls.org/issues/?address=<URL-encoded-address>
```

Maybe the result that comes back _won't_ be in a split district for your address. If that's the case we can provide a 3rd option for someone to type in their address, which is supported but I wanted to avoid having someone type from the iOS app.

We'll also have to change the messaging of that alert.

bengottlieb:
That's what I thought, but I figured it'd have a bigger effect in a place like NY than Texas, and it doesn't seem to have an effect there. I entered @jablair's data directly into the API [https://developers.google.com/civic-information/docs/v2/representatives/representativeInfoByAddress] and definitely saw more data with fewer digits. Did the same for a lat/long in Chicago and got similar results.

jablair:
@subdigital Yeah, I got back non-split district results when I entered my address. Correct combination of senators and reps, too.

I'll cop to not understanding the split district logic, but it just struck me as odd that reducing the accuracy of the location data would increase the accuracy of the lookup results. Truncating the GPS coordinates moves my location about .07 miles and neither location is close to the edge of my congressional district. It was enough to make me wonder if there was an issues dealing with the additional digits on the back end.

nickoneill:
[...]

PR #84, opened 2017-02-19
reduced lat/long accuracy to 3 signifigant digits to work around API bug #78

bengottlieb:
Looks like some weirdness in the Google Civic Information API. I think the significant digits gets us enough accuracy for anywhere in the country (from my VERY limited testing)

### DIFF
--- a/FiveCalls/FiveCalls/UserLocation.swift
+++ b/FiveCalls/FiveCalls/UserLocation.swift
@@ -59,7 +59,7 @@ class UserLocation {
     
     func setFrom(location: CLLocation, completion: @escaping (Void) -> Void) {
         locationType = .coordinates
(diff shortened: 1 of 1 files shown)


subdigital:
I don't think this is the right way to solve this particular problem.  Given the background that Nick provided on #78, I think it's probably best to augment the alert we present:

- add another `LocationType` of `.address`
- update the fetch operation to support this and URL encode it
- if we are in a split district and the `UserLocation.current.locationType == .coordinates`, then we'll have to provide them w/ a way to enter in their full address (maybe via UIAlsert style prompt?).
- maybe update the edit location screen with a button to enter in your address?

Alternatively we could just make the zip code entry a generic text field and just let the user type whatever they want there. We'd have to be better about messaging when they have malformed data, but it would support both options.

PR #85, opened 2017-02-19
Fix transparent header in issues view.

inuwali:
As noted in the comment thread for the issue, I had to add back in `IssuesContainerViewController.setContentInsets()`; its removal caused a 20-point gap to appear between the header and the table view content.


mccarron:
This looks good to me and glad it fixes the issues you were seeing. 

Gotta figure out the merging order of this with the iPad changes. @subdigital @bengottlieb what are your thoughts? Merge this in first then refactor iPad?

subdigital:
Yes, this should go in first, as we have some loose ends to work on w/ the iPad app.

FiveCalls/FiveCalls/IssuesContainerViewController.swift
```diff
@@ -38,15 +39,40 @@ class IssuesContainerViewController : UIViewController, EditLocationViewControll
         
         issuesVC.didMove(toParentViewController: self)
         issuesViewController = issuesVC
+        
+        let effectView = UIVisualEffectView(frame: headerContainer.bounds)
+        effectView.translatesAutoresizingMaskIntoConstraints = false
+        effectView.effect = UIBlurEffect(style: .light)
```
subdigital: I like to extract these into lazy properties, like this:

```swift
lazy var effectView: UIVisualEffectView = {
    let effectView = ...
    // more setup
    return effectView
}
```

Then you can just use it in `viewDidLoad` or wherever and they get set up properly.

FiveCalls/FiveCalls/IssuesContainerViewController.swift
```diff
@@ -38,15 +39,40 @@ class IssuesContainerViewController : UIViewController, EditLocationViewControll
...
+            effectView.contentView.topAnchor.constraint(equalTo: headerView.topAnchor),
+            effectView.contentView.bottomAnchor.constraint(equalTo: headerView.bottomAnchor),
+            effectView.contentView.leftAnchor.constraint(equalTo: headerView.leftAnchor),
+            effectView.contentView.rightAnchor.constraint(equalTo: headerView.rightAnchor),
+            
+            headerContainer.topAnchor.constraint(equalTo: effectView.topAnchor),
+            headerContainer.bottomAnchor.constraint(equalTo: effectView.bottomAnchor),
+            headerContainer.leftAnchor.constraint(equalTo: effectView.leftAnchor),
+            headerContainer.rightAnchor.constraint(equalTo: effectView.rightAnchor)
+            ])
```
subdigital: Should probably extract all this header setup into a method... maybe `setupBlurHeader()` ?

FiveCalls/FiveCalls/IssuesContainerViewController.swift
inuwali: I'm a Swift newbie so I had to try a few things before it worked... Needed to add () after the closing curly brace.

FiveCalls/FiveCalls/IssuesContainerViewController.swift
inuwali: Makes sense.

inuwali: BTW I was wondering why you didn't set up the container/contained relationship directly in the storyboard... Figured you had a reason but didn't investigate.

subdigital:
Thanks, @inuwali!

Issue #96, opened 2017-02-21
Marking rep as unavailable should not count as complete


armadsen:
Right now, if I mark a contact as "unavailable" it counts the same as making contact (green checkmark and all). It seems more useful for it to either show as uncontacted or yellow, or somesuch. That way, I can remember to come back to that contact later. e.g. tonight I called my rep and their voice mailbox was full. I'll try again tomorrow, but 5Calls won't remind me that I need to do that.

subdigital:
Good point. It should definitely move you to the next contact (if there is one), but not mark it as complete. Thanks for reporting this.

bengottlieb:
I put up a PR to treat these as 'completed', but then clear them the next 'day'.

PR #99, opened 2017-02-21
Reset unavailable after 12 hours

bengottlieb:
- if a call is marked as "unavailable", and is older than 12 hours, then when it's read back into the log, it's skipped
- when re-entering the app, each screen that display a log will check to see if there are any old, unavailable calls in the log, and will remove them (by reloading the log) and then reloads the visible table.

Addresses #96

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ fastlane/screenshots
 fastlane/test_output
 .env
 .ruby-version
+.DS_Store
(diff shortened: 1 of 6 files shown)


mccarron:
I think the better option is to just not treat Unavailable as Completed? 

We still want to log Unavailable and also be able to log all possible attempts over time, for the Your Impact screen, and this would remove that level of detail in the logs.

Thoughts @subdigital?

bengottlieb:
actually, yeah, that makes sense.

.gitignore
```diff
@@ -68,3 +68,4 @@ fastlane/screenshots
 fastlane/test_output
 .env
 .ruby-version
+.DS_Store
```
subdigital: Just a heads up, this belongs in your system `$HOME/.gitconfig` not the project.

subdigital:
Going to close this PR, as I think @mccarron is right. I'm still thinking about the best approach to solve this (maybe a `isCompleted` flag on `ContactLog`, but in any case it warrants a new PR.

Issue #105, opened 2017-02-25
Move to "Address or Zip Code" Entry


subdigital:
We've received a bunch of reports that niether zip code nor GPS provide 100% functionality with respect to congressional districts.  We need to allow passing a full address to the API in these cases.

- Remove zip code validation
- make text field read "Enter Zip or Address"
- update any references to zip code (we can just call this an "address" in the code)
- URL encode whatever they type and pass to API in `FetchIssuesOperation`
- update any alert dialog messaging about split districts to reflect this

subdigital:
@tomburns is working on this.

tomburns:
Welp, just noticed this :) I'll update the UI text in my PR

tomburns:
@subdigital I think this issue can be closed after PR #106

PR #106, opened 2017-02-25
Use address rather than zip for manual location entry.

tomburns:


tomburns:
Should resolve / relate to #78

tomburns:
Or #105 😛

tomburns:
re URL encoding: the Operation is already using URLComponents (:100:) so I don't think any additional escaping is necessary?

FiveCalls/FiveCalls/EditLocationViewController.swift
```diff
@@ -63,27 +64,12 @@ class EditLocationViewController : UIViewController, CLLocationManagerDelegate {
         delegate?.editLocationViewControllerDidCancel(self)
     }
     
-    @IBAction func submitZipCodeTapped(_ sender: Any) {
-        if validateZipCode() {
+    @IBAction func submitAddressTapped(_ sender: Any) {
+            //TODO: Does this event need to be renamed?
             Answers.logCustomEvent(withName:"Action: Used Zip Code")
```
subdigital: Good catch. I think yes, and if needed I can combine the values from the analytics for an overall report.

FiveCalls/FiveCalls/UserLocation.swift
```diff
@@ -50,10 +50,10 @@ class UserLocation {
...
     }
     
-    func setFrom(zipCode: String) {
-        locationType = .zipCode
-        locationValue = zipCode
-        locationDisplay = zipCode
+    func setFrom(address: String) {
+        locationType = .address
+        locationValue = address
+        locationDisplay = address
```
subdigital: Here we have to consider the case where the user had `zipCode` in user defaults already, then runs the app w/ this code later on. I think we should probably have a `Migrations` class that handles this on startup, that can manually check, something like this:

```swift
class Migrations {
  func run() {
    migrateUserDefaultsFromZipCodeToAddress()
  }

  private func migrateUserDefaultsFromZipCodeToAddress() {
     ...
  }
}
```

We can call this `run()` function early on in the `applicationDidFinishLaunching` method.

FiveCalls/FiveCalls/en.lproj/Main.strings
```diff
@@ -14,14 +14,14 @@
 /* Class = "UIButton"; normalTitle = "Set Location"; ObjectID = "6mh-1s-XbV"; */
 "6mh-1s-XbV.normalTitle" = "Set Location";
 
-/* Class = "UILabel"; text = "Or Enter Your Zip Code"; ObjectID = "7Xf-pd-kKj"; */
-"7Xf-pd-kKj.text" = "Or Enter Your Zip Code";
+/* Class = "UILabel"; text = "Or Enter Your Zip Code or Address"; ObjectID = "7Xf-pd-kKj"; */
+"7Xf-pd-kKj.text" = "Or Enter Your Zip Code or Address";
```
subdigital: I think this would read better as:

`Or enter manually:`

(Can you also increase the spacing between the "Use My Location" button and this label?)

iphone 7 ios 10 2 14c89 2017-02-25 14-19-09

FiveCalls/FiveCalls/UserLocation.swift
tomburns: Addressed with a custom rawValue init on the location type enum

FiveCalls/FiveCalls/en.lproj/Main.strings
tomburns: done 📏

subdigital:
This resolves our biggest complaint in the reviews, so thank you! 🙏

PR #107, opened 2017-02-25
Don’t count “unavailable” as completed.

tomburns:
This is my naive stab at #96; let me know if this doesn't make sense as a place for this revised logic.


subdigital:
I think this is reasonable for now.  Thanks!

PR #108, opened 2017-02-25
Fixed i phone 7 plus rotation

bengottlieb:
fixed split view on the iPhone n+


subdigital:
Ah, I wonder if this was a project file merge issue...

bengottlieb:
I'm not sure I follow; on the 6+, this is only used when in landscape (when in splitview)

PR #109, opened 2017-02-25
Add an instructive alert before placing first call

### LINKED ISSUE #90: Messaging for first call
I'd like to add a one-time alert when the user calls to give them some quick instructions about the need to double-tap & come back so they can read the script.

First time user calls, present an alert:

> We're about to place the call. When it starts dialing, turn on Speaker phone, then double tap the home button come back to this app so you can read the script.
>
> **OK, Got it** | Cancel

When we press this button:

- Save an entry in user defaults so we know not to present this alert again
- Call the system dialer

tomburns:
Resolves #90


subdigital:
And that heads off the 2nd most popular feedback item we're receiving.  @tomburns is on 🔥

Issue #114, opened 2017-02-28
Allow copying script


subdigital:
Someone requested this, perhaps so they could tweak the script a little when the called.

PR #115, opened 2017-03-01
Display script in a UITextView, enabling copying

tomburns:
#114 - Editing and scrolling disabled, and I transferred the styling from the previous label; seems to do the trick

simulator screen shot feb 28 2017 10 56 10 pm


mccarron:
Thanks that's perfect!

PR #120, opened 2017-03-08
Move pantry files to app support

subdigital:
This checks on app startup for existence of the pantry cache folder, migrating it to Application Support if necessary. If the target folder exists already or some other error happens, we log it and continue.

The user will lose data if this ever happens, but it would be pretty rare and not entirely critical.

(Worth noting that users with low disk space are already losing data :/)


bengottlieb:
Looks good to me.

Issue #121, opened 2017-03-08
Gracefully handle office locations w/ no names


subdigital:
Sometimes the office names aren't present in the API response for the House.  Just need to check for this so we don't see something like:  `"222-333-4567 ()"`

bengottlieb:
Do we have any examples?

subdigital:
It's dependent on the data, but try to find an issue with a house representative.  I found one w/ zip 77429 and I posted the picture to slack.

bengottlieb:
Hmm, every issue I'm seeing is either Senate- or Governor-only, or just hits a House subcommittee.

mccarron:
Fixed by #122

PR #122, opened 2017-03-08
Fix for #121 (office locations with no name)

bengottlieb:


mccarron:
Looks good. Tagging issue #121

PR #129, opened 2017-03-20
Reminder bell on main screen

subdigital:
Based on work by @chrisbrandow, this adds an icon for the bell in favor of the 🔔 emoji. This also gives us the ability to show selected states based on whether reminders are enabled or not.

reminder-bell


nickoneill:
:+1: this looks really good!

Issue #134, opened 2017-03-28
Change font from Roboto to the iOS System Font


mergesort:
We currently use a UIApperance proxy to assign the [Roboto Condensed font](https://github.com/5calls/ios/blob/b01ed57d0e65a2212bfe6a8956d31c9d4e2b6291/FiveCalls/FiveCalls/Appearance.swift#L16) throughout the app.

I propose we switch from Roboto to San Francisco, or using the system default, so that we match our platform rather than maintaining parity across platforms.

tomburns:
I'll have some time this week to grab this if we decide to proceed - we can probably do it in a couple phases, first ripping out the custom font and second cutting over to dynamic type where possible throughout the app.

mergesort:
I was thinking if @mb is interested, this would be a good starter task for him. Thoughts?

mergesort:
I was thinking if @mattbischoff is interested, this would be a good starter task for him. Thoughts?

tomburns:
@mergesort fine by me! just wanted to get the ball rolling, but I can cut over to something else instead.

mattbischoff:
I am!

subdigital:
I was in the middle of a related change and somehow got sucked into doing this whole thing. Might need some tweaking here or there, but currently uses dynamic type supported fonts everywhere except nav bar title and tableview section headers.

PR #136, opened 2017-03-29
Remove Custom Font

tomburns:
An initial stab at #134 

If we go this way, I'm happy to continue before we merge this by adding dynamic type support to the high traffic screens etc where applicable.


tomburns:
This should probably be considered a WIP PR until I've done a pass for dynamic type, etc.

mccarron:
Would love so see some screenshots as you progress.

tomburns:
Closing this for now as @mattbischoff is going to take it over 👍

Issue #145, opened 2017-04-02
Inform users about fax/email situation on script copy


nickoneill:
We initially added script copying because some users wanted to print out scripts when they call. Now more requests are coming through to add copying so people can engage on faxes or emails which we'd like to discourage - not because we're greedy but because we believe calls are genuinely the best way to make an impact.

When users select text on the scripts to copy them, we should display some sort of informational message indicating that calls are more effective than faxing or emailing. Copy TBD.

bengottlieb:
I just pushed a PR for this. Might want to change it to only trigger if they copy more than, say, 90% of the text? Alert only fires once.

PR #146, opened 2017-04-02
added an alert when copying script text about the dangers therein for #145

bengottlieb:


FiveCalls/FiveCalls/Localizable.strings
```diff
@@ -47,3 +47,5 @@
 "scheduled-reminders-description"     = "Turn these on to get a quick local reminder to make your 5 calls.";
 "line-busy"                           = "Line Busy? ▼";
 "choose-a-number"                     = "Select a Different Office";
+"think-before-copying-alert-title"    = "Please Consider Before Texting, Faxing, or Emailing";
```
mccarron: Missing "Calling" here after "Consider"?

FiveCalls/FiveCalls/IssueScriptCell.swift
```diff
@@ -14,3 +14,18 @@ class IssueScriptCell: UITableViewCell {
...
+    
+    override func copy(_ sender: Any?) {
+        if !UserDefaults.standard.bool(forKey: UserDefaultsKeys.hasWarnedAboutDangersOfCopying.rawValue) {
+            let alert = UIAlertController(title: R.string.localizable.thinkBeforeCopyingAlertTitle(), message: R.string.localizable.thinkBeforeCopyingAlertBody(), preferredStyle: .alert)
+            
+            alert.addAction(UIAlertAction(title:  R.string.localizable.okButtonTitle(), style: .cancel, handler: { action in }))
+            self.viewController?.present(alert, animated: true, completion: nil)
+            UserDefaults.standard.set(true, forKey: UserDefaultsKeys.hasWarnedAboutDangersOfCopying.rawValue)
+        }
+        super.copy(sender)
```
mccarron: Should we maybe call `super` first?

bengottlieb: why?

FiveCalls/FiveCalls/Localizable.strings
bengottlieb: I think the meaning is more like "Think about", rather than a specific option. I figure, we're in an app called "5 Calls". Any consideration would kinda have to take Calling into account.

bengottlieb:
Here's the alert:
simulator screen shot apr 11 2017 10 32 25 pm

nickoneill:
Can we do a break in there? Slight copy change suggestions:

> Calling has been consistently shown to be the most effective way to make your voice heard with your Representatives.
> 
> Please consider this before using another method of communication.

bengottlieb:
Copy change pushed.

bengottlieb:
simulator screen shot apr 11 2017 10 32 25 pm

bengottlieb:
simulator screen shot apr 12 2017 8 24 11 am

mccarron:
Thanks, looks good with that break. Merging.

PR #149, opened 2017-04-11
Made operation a weak, rather than an unowned, argument

bengottlieb:


mccarron:
This does appear to fix the crash I could reproduce, I no longer can. Thanks @bengottlieb.

Issue #143, opened 2017-03-29
Submit button should be disabled until location text is entered.


subdigital:

keithgee:
Pull request #151 disables the submit button when the location field is empty, and also adds transparency to the BlueButtons when they are disabled.

PR #151, opened 2017-04-17
Disable Submit button when location is empty

keithgee:
With this change, the Submit button in the edit location view is enabled only
when there is text in the location textfield.

This addresses issue #143 in the 5calls/ios issue list.

### DIFF
--- a/FiveCalls/FiveCalls/BlueButton.swift
+++ b/FiveCalls/FiveCalls/BlueButton.swift
@@ -47,14 +47,12 @@ class BlueButton: UIButton {
 
     override var isHighlighted: Bool {
         didSet {
-            titleLabel?.alpha = 1.0
             updateColors()
         }
     }
     
     override var isSelected: Bool {
         didSet {
-            titleLabel?.alpha = 1.0
             updateColors()
         }
     }
@@ -70,17 +68,11 @@ class BlueButton: UIButton {
         else if isSelected { backgroundColor = selectedBackgroundColor }
         else { backgroundColor = normalBackgroundColor }
         
-        if isEnabled {
-            alpha = 1.0
-            titleLabel?.alpha = 1.0
-        }
-        else {
-            alpha = 0.4
-            titleLabel?.alpha = 0.4
-        }
-
         setTitleColor(.white, for: .selected)
         setTitleColor(.white, for: .highlighted)
         setTitleColor(defaultTextColor, for: .normal)
+        
+        titleLabel?.alpha = 1.0
+        alpha = isEnabled ? 1.0 : 0.5
     }
 }
--- a/FiveCalls/FiveCalls/BlueButton.swift
+++ b/FiveCalls/FiveCalls/BlueButton.swift
@@ -59,10 +59,25 @@ class BlueButton: UIButton {
         }
     }
 
+    override var isEnabled: Bool {
+        didSet {
+            updateColors()
+        }
+    }
+    
     func updateColors() {
         if isHighlighted { backgroundColor = highlightBackgroundColor }
         else if isSelected { backgroundColor = selectedBackgroundColor }
         else { backgroundColor = normalBackgroundColor }
+        
+        if isEnabled {
+            alpha = 1.0
+            titleLabel?.alpha = 1.0
+        }
+        else {
+            alpha = 0.4
+            titleLabel?.alpha = 0.4
+        }
 
         setTitleColor(.white, for: .selected)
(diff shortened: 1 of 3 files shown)


nickoneill:
Thanks @keithgee! It's the subtle things that get under your skin to the point that you just need to fix them yourself, right? 😂 We'll review shortly

FiveCalls/FiveCalls/EditLocationViewController.swift
```diff
@@ -37,6 +38,7 @@ class EditLocationViewController : UIViewController, CLLocationManagerDelegate {
         
         if case .address? = UserLocation.current.locationType {
```
mccarron: I think before we enable the button, we should also validate `UserLocation.current.locationValue` has anything. My simulator has the legacy location data so it returns like so:

```type: address value:  | display: My Location```

Which causes the button to be enabled with nothing in it. 

simulator screen shot apr 17 2017 8 47 48 pm

Maybe call `addressTextFieldChanged` so that we're changing the `isEnabled` in one place?

File: FiveCalls/FiveCalls/BlueButton.swift [hunk 2]
```diff
@@ -59,10 +59,25 @@ class BlueButton: UIButton {
...
+            updateColors()
+        }
+    }
+    
     func updateColors() {
         if isHighlighted { backgroundColor = highlightBackgroundColor }
         else if isSelected { backgroundColor = selectedBackgroundColor }
         else { backgroundColor = normalBackgroundColor }
+        
+        if isEnabled {
```
mccarron: Can we maybe re-factor this to handle the `titleLabel?.alpha` rules from `isHighlighted` and `isSelected` as well?

FiveCalls/FiveCalls/BlueButton.swift
```diff
@@ -59,10 +59,25 @@ class BlueButton: UIButton {
...
+            alpha = 1.0
+            titleLabel?.alpha = 1.0
+        }
```
mccarron: Can we put this on one line? 
```} else {```

File: FiveCalls/FiveCalls/EditLocationViewController.swift [hunk 4]
```diff
@@ -70,8 +72,14 @@ class EditLocationViewController : UIViewController, CLLocationManagerDelegate {
...
     }
+    
+    @IBAction func addressTextFieldChanged() {
+        if let addressLength = addressTextField.text?.characters.count {
+            submitButton.isEnabled = addressLength > 0
+        }
+    }
 
-    //Mark: CLLocationManagerDelegate methods
+    //MARK: - CLLocationManagerDelegate methods
```
mccarron: 👍

keithgee:
Thanks, @mccarron. I have to finish filing taxes today, but I'll work on this soon. Do you want me to squash the changes into the original commit before I push them up?

@nickoneill, I heard part of a podcast with @subdigital talking about the community effort behind the app. When I saw the starter-issue tags in the list, I became encouraged about the chance to contribute to a worthwhile project and to meet iOS developers.

subdigital:
@keithgee awesome!

mccarron:
That's great! Happy to have you! 

@keithgee you don't have to squash if it's too much for you, as I always squash on merge. Also no rush, finish those taxes first!

FiveCalls/FiveCalls/EditLocationViewController.swift
keithgee: I've updated the code to call addressTextFieldChanged, instead of setting isEnabled directly. Also, addressTextFieldChanged makes sure there is something other than white space in the field before enabling the button.

FiveCalls/FiveCalls/BlueButton.swift
keithgee: Refactored. Also, I found that when I lower the alpha level to make the button appear disabled, that also affects the appearance of the label automatically.

The code to set the alpha level of the label to 1.0 is now used every time updateColors is called, but it's not clear to me why this was needed in isSelected and isHighlighted. I see the same behavior when I remove that line of code.

FiveCalls/FiveCalls/BlueButton.swift
keithgee: After updating, I was able to manage like so:
alpha = isEnabled ? 1.0 : 0.5

mccarron:
Thanks for the updates, looks great now. Merging in!

PR #154, opened 2017-04-17
Changed tint color for UINavigationController

musicisamiracle:
This changes the tint color for the UINavigationController in the About storyboard to white. The original tint was the system blue color which did not contrast well with the blue color of the navigation bar.


nickoneill:
Thanks @musicisamiracle! What elements does this change the color on so I can test?

musicisamiracle:
It appears to only change the back button that is created by the UINavigationController when you navigate through the view controllers in the About.Storyboard.

mccarron:
Ahh good catch! Merging this in.

Issue #153, opened 2017-04-17
iPad only: Change in user location does not immediately update representative list


keithgee:
In the views for the iPad app, the user can set a location with a pop-up style window while viewing an issue. This is convenient.

If the user changes location in the "Set your location" window, ideally the list of representatives shown at the bottom of the issue view would update immediately to reflect the new location. In the current version of the iPad app, the representatives in the "Call your representatives" list reflect the previous location until a new issue is tapped.

erikkrietsch:
I can take this one.

Posting this mostly for my own future self: I know _why_ this is happening, but I haven't decided yet how I would want to fix it.  The `IssueDetailViewController` is setting itself as the delegate for the `EditLocationViewController` when it is launched via a segue on the `IssueDetailViewController` but that only happens when the device is a particular size. Subscribing to notifications might be better, but will have to be reconciled with the delegate code so the two aren't stepping on each other.

mccarron:
@erikkrietsch Sounds like we need to set the delegate in the other case maybe when the popup is launched? 

If that seems too hard to do properly, I'd be game for listening for a notification across the app for this. Will make it easier going forward to adding setting location from other places in the future.

erikkrietsch:
There actually already is a notification for when a user changes their location, so I was going to just tap into that. I've got the fix just about ready, I just need to extract out a common function for the two scenarios.

erikkrietsch:
It's actually a little more complicated than I initially thought, and my original assessment wasn't quite right. If the user hasn't yet set their location, then they can do that via a special table cell in the IssueDetailViewController. That's when the IDVC sets itself as the delegate for the location controller. This can happen on any size device. The iPad can also set the location through a specially added button to the header in the navigation controller, but the IDVC has no direct interaction with it. Subscribing to the notification works well in this case, but in the prior scenario, we double dip and wind up hitting the network twice—once for the delegate call and once for the notification. 

I'm still working on it. I think what I'm going to do is separate out the work that was happening on the delegate call and move the networking and UI refresh into the notification handler.

PR #155, opened 2017-04-24
Refresh contacts on location change with iPad. Addresses #153.

erikkrietsch:
I couldn't separate the logic like I mentioned doing in issue #153, so I wound up removing the observer when setting the location via `EditLocationViewController` then add it back once that action is completed.

### DIFF
--- a/FiveCalls/FiveCalls/IssueDetailViewController.swift
+++ b/FiveCalls/FiveCalls/IssueDetailViewController.swift
@@ -15,7 +15,9 @@ class IssueDetailViewController : UIViewController, IssueShareable {
     var issuesManager: IssuesManager!
     var issue: Issue!
     var logs: ContactLogs?
-    
+
+    var shouldAddLocationObserver: Bool = false
+
     @IBOutlet weak var tableView: UITableView!
     
     override func viewDidLoad() {
@@ -26,6 +28,7 @@ class IssueDetailViewController : UIViewController, IssueShareable {
         tableView.estimatedRowHeight = 100
         tableView.rowHeight = UITableViewAutomaticDimension
         NotificationCenter.default.addObserver(self, selector: #selector(madeCall), name: .callMade, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(changedLocation), name: .locationChanged, object: nil)
     }
 
     func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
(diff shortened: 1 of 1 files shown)


FiveCalls/FiveCalls/IssueDetailViewController.swift
```diff
@@ -56,6 +59,36 @@ class IssueDetailViewController : UIViewController, IssueShareable {
         tableView.reloadRows(at: tableView.indexPathsForVisibleRows ?? [], with: .none)
     }
 
+    func changedLocation() {
+      didUpdateLocation()
+    }
+
+    func didUpdateLocation(using viewController: UIViewController? = nil) {
+        issuesManager.fetchIssues { result in
+            if let vc = viewController, self.issuesManager.isSplitDistrict {
```
mccarron: I think we should `[weak self]` these self references.

FiveCalls/FiveCalls/IssueDetailViewController.swift
```diff
@@ -56,6 +59,36 @@ class IssueDetailViewController : UIViewController, IssueShareable {
...
+            } else {
+                viewController?.dismiss(animated: true, completion: nil)
+            }
+
+            if let issue = self.issuesManager.issue(withId: self.issue.id) {
+                self.issue = issue
+                self.tableView.reloadSections([IssueSections.contacts.rawValue], with: .automatic)
+                self.scrollToBottom()
+            } else {
+                // weird state to be in, but the issue we're looking at
```
mccarron: Thanks for the comment helpful for looking at this later.

subdigital:
Thanks for putting this together, and yeah... yikes, this isn't ideal.

My first thought is that the issues container vc is reaching directly into the issues vc to tell it to `loadIssues()`, and this is the crux of the issue.

There are multiple visible parties who need to respond to this event, so I think this should be reworked.

So what if we moved that responsibility up into the IssuesContainerViewController, and once it is done it can send a notification that issues were loaded?

Issues VC then responds by updating it's array / table view, and the detail VC responds by pulling out the relevant issue and rebinding the UI.

There might be some kinks we'll run into implementing this that I'm not seeing, but I think this is probably better than turning on/off notification handlers.

Thoughts welcome.

erikkrietsch:
Yeah, certainly agreed that this solution isn’t ideal. 

I had some similar thoughts, although my initial instinct was to get any location responsibility up out of the `IssueDetailViewController` and probably push that into `EditLocationViewController`. 

To be honest, I wasn’t sure how much leeway I had to operate, and didn’t want to overstep the bounds of this particular issue. But since you’ve broached the subject, I’ll consider a more holistic solution.

You may also be right that the `IssuesContainerViewController` can handle these responsibilities and push issues into the detail controller, though I’m not sure how the iPhone layout would respond when the user is in a split district. I’ll try out a couple things and see what I can come up with.

subdigital:
Completely understand about not wanting to increase the scope of the issue. I think it's worth it to improve some of these awkward interactions. Thanks for taking the time to look into this! 🙏

Issue #152, opened 2017-04-17
Pressing return on location page should hide keyboard and submit location


keithgee:
On the screen to edit the user location, pressing the return key on the keyboard does nothing. It should dismiss the keyboard if there is text in the field, and probably should automatically submit the updated location as well.

This would be a good starter-issue.

PR #157, opened 2017-04-25
Pressing return submits location from addressTextField. Addresses #152

musicisamiracle:
In the Edit Location View Controller, nothing would happen if you tapped "Done" on the keyboard when entering your location manually. Now, it will resign the keyboard and submit the location.


ianyh:
Referencing the issue in a comment to get it linked on there: #152

bryanjclark:
This looks great to me!

Issue #159, opened 2017-04-30
Refresh issues list periodically


nickoneill:
Some (probably less-tech savvy) people email us and say that the issues list doesn't refresh day-to-day. I suspect this is older users who infrequently use their device (sometimes iPads) so the app isn't getting killed between launches, leading to out-of-date issues.

Two things we could do here:
* Provide automatic refresh when the app is foregrounded if the issues haven't been refreshed in ~12 hours
* Pull-to-refresh control on the issues list for manually checking for new issues

mccarron:
Closed by #163

PR #163, opened 2017-05-05
Refresh issues outside of just app launch

ianyh:
Fix for #159. I went for reloading on every launch because it seemed logically simpler, but if there's a good reason to space it out I can change it.


mccarron:
This looks like a great low impact fix, thanks for it.

Issue #148, opened 2017-04-11
Support "More issues" list


nickoneill:
We've been offering still-valid-yet-not-frontpage issues on our "more" page: https://5calls.org/more

This should be supported if users want to see issues beyond what the front page has to offer.

Keep in mind this will be supplemented with issue categories in the near future: some section headers on the table views are probably enough for that.

piebob:
i'd suggest this is fairly high priority--we get a lot of people emailing who don't know there are a whole bunch more issues.

PR #164, opened 2017-05-12
Add an option for displaying all issues

ianyh:
Fix for #148 

General approach was to make the issues view controller polymorphic, and then use the table footer view to deal with the transition. Screenshots below.

simulator screen shot may 11 2017 8 07 52 pm
simulator screen shot may 11 2017 8 07 56 pm

### DIFF
--- a/FiveCalls/FiveCalls/FetchIssuesOperation.swift
+++ b/FiveCalls/FiveCalls/FetchIssuesOperation.swift
@@ -11,6 +11,7 @@ import Foundation
 class FetchIssuesOperation : BaseOperation {
     
     let location: UserLocation?
+    let shouldFetchAllIssues: Bool
     
     // Output properties.
     // Once the job has finished consumers can check one or more of these for values.
@@ -18,8 +19,9 @@ class FetchIssuesOperation : BaseOperation {
     var error: Error?
     var issuesList: IssuesList?
 
-    init(location: UserLocation?) {
+    init(location: UserLocation?, shouldFetchAllIssues: Bool) {
         self.location = location
+        self.shouldFetchAllIssues = shouldFetchAllIssues
     }
     
     lazy var sessionConfiguration = URLSessionConfiguration.default
@@ -41,6 +43,10 @@ class FetchIssuesOperation : BaseOperation {
     }
     
     func addressQueryString() -> String? {
+        guard !shouldFetchAllIssues else {
+            return "inactive=true"
+        }
+
         guard let location = self.location,
               let value = location.locationValue
             else { return nil }
--- a/FiveCalls/FiveCalls/FetchIssuesOperation.swift
+++ b/FiveCalls/FiveCalls/FetchIssuesOperation.swift
@@ -8,10 +8,20 @@
 
 import Foundation
 
+fileprivate extension IssuesManager.Query {
+    func queryString() -> String? {
+        switch self {
+        case .all:
+            return "inactive=true"
+        case .top(let location):
+            return location?.locationValue.flatMap { "address=\($0)" }
+        }
+    }
+}
+
 class FetchIssuesOperation : BaseOperation {
     
-    let location: UserLocation?
-    let shouldFetchAllIssues: Bool
+    let query: IssuesManager.Query
     
     // Output properties.
     // Once the job has finished consumers can check one or more of these for values.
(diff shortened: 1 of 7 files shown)


mccarron:
Great implementation. Clean and re-uses a lot of what we have already. 

Could you make the table footer color match the rest of the rows? It seems too harsh in that flat white. Maybe a slightly larger font too?

FiveCalls/FiveCalls/FetchIssuesOperation.swift
```diff
@@ -11,15 +11,17 @@ import Foundation
...
+    let shouldFetchAllIssues: Bool
     
     // Output properties.
     // Once the job has finished consumers can check one or more of these for values.
     var httpResponse: HTTPURLResponse?
     var error: Error?
     var issuesList: IssuesList?
 
-    init(location: UserLocation?) {
+    init(location: UserLocation?, shouldFetchAllIssues: Bool) {
```
subdigital: Not sure I'm a fan of a boolean for this. Something a little less prescriptive (from the caller's perspective) might look like this:

```swift
enum IssueQuery {
    case all
    case top
}
```

Then the constructor could accept this type instead, so we'd call it like this:

```swift
FetchIssuesOperation(userLocation, query: .top)

// or

FetchIssuesOperation(userLocation, query: .all)
```

Going even further, perhaps we could move the URL creating into the enum, so the `.all` case would append the `inactive=true` query param.

Not sure if the last part fits, but just spitballing.

Doing this would give us flexibility to add more query variations in the future as well without modifying the constructor parameters.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -72,12 +79,15 @@ class IssuesViewController : UITableViewController {
     }
     
     override var preferredStatusBarStyle: UIStatusBarStyle {
-        return .lightContent
+        return shouldFetchAllIssues ? .default : .lightContent
```
subdigital: We can remove this entire method. The initial VC is actually the `IssuesContainerViewController`, which contains the blurred header thing. This VC doesn't override that method, so it's `.default` anyway.

When we push the more issues instance, the navigation controller will ask the top VC for the style, but it would be `.default` anyway.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -16,6 +16,13 @@ protocol IssuesViewControllerDelegate : class {
 }
 
 class IssuesViewController : UITableViewController {
+
+    @IBInspectable var shouldFetchAllIssues: Bool = false {
+        didSet {
+            issuesManager.shouldFetchAllIssues = true
```
subdigital: I'm not a fan of setting a property and then later firing a method, and the method's behavior is dependent on this (especially since these two steps are separated physically in the code).

I'd like to move the logic into the `fetchIssues` method directly.  We can leverage the same enum for this.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -72,12 +79,15 @@ class IssuesViewController : UITableViewController {
...
-        return .lightContent
+        return shouldFetchAllIssues ? .default : .lightContent
     }
     
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         logs = ContactLogs.load()
+        if shouldFetchAllIssues {
+            navigationController?.setNavigationBarHidden(false, animated: animated)
+        }
```
subdigital: I'm tempted to suggest a subclass for this.  The `AllIssuesViewController` could have all of these methods defined as simple overrides, including the fetch call.

@mccarron thoughts on this?

subdigital:
@ianyh Thanks so much for putting this together. I've been really busy with other things and haven't been able to carve out much time for 5 Calls this week. Let me know if you want to pair on some of the above changes, I'd be game for it.

ianyh:
Changed the issues manager to take a query as part of the fetch call, which encapsulates the polymorphic behavior. It does feel like a cleaner interface now.

I also changed the button to be a cell, which is styled closer to the other cells. I tried styling the button itself and still using the footer view, but it didn't look right without the separator. Screenshot below.

simulator screen shot may 12 2017 11 06 09 pm

FiveCalls/FiveCalls/IssuesViewController.swift
ianyh: I considered this and opted against it. It felt unnecessarily heavy-weight. I am generally pretty averse to inheritance, though.

PR #166, opened 2017-07-20
Move UI of issues list closer to iOS visual design

ianyh:
I've done some work on this, and it is probably at the stage where it could use eyes.


subdigital:
Looking good so far. I noticed a few issues here and there. I think we're close enough to push forward with this implementation.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -125,7 +156,16 @@ class IssuesViewController : UITableViewController {
         loadIssues()
     }
 
+    fileprivate func setTitleLabel(location: UserLocation?) {
+        locationButtonItem?.title = location?.locationDisplay ?? "Set Location"
```
subdigital: Need localization for "Set Location" if it doesn't exist already.  (Check `R.strings.localized.`...)

subdigital:
This button color looks like these are disabled. The same color is on the welcome screen.

nickoneill:
It didn't look like those button colors were part of this commit, I changed them to a darker blue in #169

ianyh:
I do plan on finishing this, FYI. I haven't quite had the spoons lately to get it across the line, but I'll hopefully get to it soon.

FiveCalls/FiveCalls/IssuesViewController.swift
ianyh: Done. There was already a localizable string for it!

ianyh:
Turns out I had the spoons. I fixed the issue with iPad first launch. I figured I would hold off on rebasing from other changes until the new commits are reviewed, and the relevant PRs are merged.

nickoneill:
Thanks! Working on getting the other items in shortly, I'll check this out as well

ianyh:
Okay, it's been rebased onto master. I think the storyboard changes from #167 are still correct.

nickoneill:
Thanks @ianyh! I'll check this out

PR #167, opened 2017-07-27
Fixes issue detail and contacts in more issues

nickoneill:
- Storyboard connections for getting from the more issues list -> issue detail weren't set
- Both main and moreissues/inactive issue endpoint needs a location sent to get contacts back

### DIFF
--- a/FiveCalls/FiveCalls/FetchIssuesOperation.swift
+++ b/FiveCalls/FiveCalls/FetchIssuesOperation.swift
@@ -10,7 +10,7 @@ import Foundation
 
 class FetchIssuesOperation : BaseOperation {
 
-    let fetchInactive: Bool
+    let query: IssuesManager.Query
     let location: UserLocation?
     
     // Output properties.
@@ -19,8 +19,8 @@ class FetchIssuesOperation : BaseOperation {
     var error: Error?
     var issuesList: IssuesList?
 
-    init(inactive: Bool, location: UserLocation?) {
-        self.fetchInactive = inactive
+    init(query: IssuesManager.Query, location: UserLocation?) {
+        self.query = query
         self.location = location
     }
     
@@ -36,7 +36,7 @@ class FetchIssuesOperation : BaseOperation {
         urlComponents.path = "/issues/"
 
         var queryItems: [URLQueryItem] = []
-        if fetchInactive {
+        if query == .inactive {
             let inactiveQueryItem = URLQueryItem(name: "inactive", value: "true")
             queryItems.append(inactiveQueryItem)
         }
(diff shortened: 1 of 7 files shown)


FiveCalls/FiveCalls/FetchIssuesOperation.swift
```diff
@@ -43,11 +34,18 @@ class FetchIssuesOperation : BaseOperation {
         urlComponents.scheme = "https"
         urlComponents.host = "5calls.org"
         urlComponents.path = "/issues/"
-        
-        if let queryString = query.queryString() {
-            urlComponents.query = queryString
+
+        var queryItems: [URLQueryItem] = []
+        if fetchInactive {
+            let inactiveQueryItem = URLQueryItem(name: "inactive", value: "true")
```
subdigital: TIL...

FiveCalls/FiveCalls/IssueDetailViewController.swift
```diff
@@ -196,7 +196,7 @@ extension IssueDetailViewController : EditLocationViewControllerDelegate {
     }
     
     func editLocationViewController(_ vc: EditLocationViewController, didUpdateLocation location: UserLocation) {
-        issuesManager.fetchIssues(.top(location)) { result in
+        issuesManager.fetchIssues(false, location: location) { result in
```
subdigital: I think I like the enum style better, but I see the need to extract the location and pass that regardless.  Maybe just remove the associated type and use the enum like this?

```swift
fetchIssues(.all, location: location)  // or .top
```

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -105,8 +105,8 @@ class IssuesViewController : UITableViewController {
         isLoading = true
         tableView.reloadEmptyDataSet()
         issuesDelegate?.didStartLoadingIssues()
-        let query: IssuesManager.Query = shouldFetchAllIssues ? .all : .top(UserLocation.current)
```
subdigital: and this line could probably have been improved as a separate method that returns the appropriate `Query` case.

subdigital:
Note for @ianyh, you probably want to rebase this branch once Nick is done.

ianyh:
Okay. 👍

FiveCalls/FiveCalls/IssueDetailViewController.swift
nickoneill: My first attempt was just both cases of `Query` had the `UserLocation` associated type. Was easy to swap to that, but it feels funky. I'll go back to the plain enum here.

FiveCalls/FiveCalls/IssuesViewController.swift
nickoneill: Based on this `IBInspectable` thing? Not sure how that would work

nickoneill:
OK, this should be set with the requested changes.

nickoneill:
ping @subdigital

PR #181, opened 2017-10-07
Swift 4 migration

### LINKED ISSUE #179: Update the project to Swift 4
I'm working on a basic migration - PR to follow.

Abizern:
This is a bare-bones migration to Swift 4 and Xcode 9 that fixes #179 

**Done**:

- Changed Swift version to 4
- Updated Pods
- Changed deployment target to 10.2 to match the original project settings
- Fixed a minor issue that was causing Tests not to build
- A couple of minor changes to reduce the original number of build warnings

**Not Done**:
- These are the minimal amount of reasonable changes to support the migration. I haven't audited the code to see where more Swifty code can be used.

**Issues**:

- UILocalNotification deprecation warnings aplenty, we should move to UN User Notifications.
- A few autolayout warnings.
- UITest failure.
- Build warning for documentation from CPDAcknowledgements. Not something that can be fixed quickly.

Subject to comments on this PR I propose to move the Issues to their own actual issues.


FiveCalls/FiveCallsTests/ContactLogsTests.swift
```diff
@@ -31,7 +31,7 @@ class ContactLogsTests: XCTestCase {
         let dateFormatter = DateFormatter()
         dateFormatter.dateFormat = "yyyy-MM-dd"
         let date = dateFormatter.date(from: "1984-01-24")!
-        let log = ContactLog(issueId: "issue1", contactId: "contact1", phone: "111-222-3333", outcome: .voicemail, date: date)
+        let log = ContactLog(issueId: "issue1", contactId: "contact1", phone: "111-222-3333", outcome: R.string.localizable.outcomesVoicemail(), date: date)
```
nickoneill: This should not be localized, it's a key sent from the server. Can replace here with 'voicemail'.

nickoneill: Or maybe I don't understand how this is generated in R.swift? I'm not clear what this will actually return.

nickoneill:
Looks pretty straightforward, just the one comment on the fix for the test (my fault for not migrating it when I worked on custom outcomes).

Agreed on the other warnings you're seeing being separate issues. UserNotifications seems like the highest priority, it might conflict with the existing WIP PR for remote notifications (but not big deal, that PR is some very basic changes i.e. the getting started section for OneSignal).

Feel free to propose any other Swift refactors as you mentioned, @subdigital is probably best to judge there.

Abizern:
Yep, you're right about the localisation. Used the actual string instead.

Also updated the project gems and pods again, as this seems a good time to do it.

nickoneill:
Not sure if you have the ability to click the merge button here when you're ready... if not, let me know if you're all set here.

Abizern:
Unfortunately, I have no Merge button. But I'm done, so merge at will. Thanks.

nickoneill:
Awesome, thanks!

Issue #138, opened 2017-03-29
Page between contacts


subdigital:
Instead of pushing a separate view controller when a call is made, it would be preferable to embed each call script in a page view controller and allow paging between them.  Tapping the completion buttons would just change the page index.

Bonus here is a custom page indicator that indicated the status of contact.  Green checkmark for _contacted_, empty circle or dot for _not yet contacted_ and some other indication of what the current page is (maybe a ring around the active one)

PR #186, opened 2017-10-09
Native review prompt

### LINKED ISSUE #173: Implement native review prompt
We should ask the user to leave us a review using the native review prompt after the user has completed x number of calls

### LINKED ISSUE #187: Use the native SKStoreProductViewController for when a user manually wants to leave a review.
In the About page of the app, one of the options is to leave a review. This creates URL than is opened in Safari.

Another option would be to push an SKStoreProductViewController with the iTunes app id, which would open the view within the app rather than externally to the app.

Just a thought, comments are welcome.

Abizern:
Fixes #173 and fixes #187 

After the end of the call list has been reached 5 times a native request for a review will be made.

Changes to the way the list of representatives is shown #138 will make this a little more robust.

The number is tracked in the user defaults, and reset when an app update is installed. Since the system takes care of deciding when it is relevant to show the request this doesn't mean that an undue number of requests will be seen.

Also, because the system decides when to show the prompt or not, this is not something that can be tracked. It was my original thought, but I don't think it will work.


Abizern:
Updated to use the same method for requesting a review from the AboutViewController for supported iOS versions.

nickoneill:
Wow, I didn't realize the rating prompt was such a minimal API. No text customization or anything.

This looks great though - good call getting rid of `fvc_open` now that it's 10+

PR #191, opened 2017-10-11
No inset adjustments in ios 11 (to be revised later)

nickoneill:
+ testing a circle ci config


nickoneill:
Circle CI runs tests too!

nickoneill:
We'll handle the actual inset issue here when we revise the header for iPhone X. In the meantime, turn off the inset adjustments in iOS 11.

Issue #174, opened 2017-09-23
Group by category on the more issues page


nickoneill:
We already send category names along with the issues on "more issues", these should be organized by their category with a nice little header.

(may want to consider using the new `all=true` query param which fetches all issues at once rather than active, then inactive later)

indraShan:
@nickoneill 
First timer, so may be a stupid question. If I understand the issue correctly: on 'more issues' screen (after tapping 'more issues' cell from Home) we want to group issues by their category, is this correct? But on Home we don't want any changes (i.e. continue showing all issues together like the app does now).

Currently category is not even parsed by the app (it is being sent by the server though.) So to close this bug: Make changes to the Issue struct to parse and save category of the issue. Then make UI changes to add a 'header' cell for every section (sections will be created based on how many unique categories we have received in the response). Is my understanding correct?

nickoneill:
Exactly correct @indraShan. Categories are attached to each issue so you'll have to look through to see which categories exist, then organize the issues into groups based on what you have.

If you're gonna take a shot at this, just mention it here so anyone else who stops by knows it's being worked on.

indraShan:
@nickoneill I will give it a shot.

nickoneill:
Awesome, let me know if you have any questions.

nickoneill:
fixed in #194

PR #194, opened 2017-10-15
- Changes to group issues by categories on the 'more issues' screen. …

indraShan:
#174 - Group by category on the more issues page

Hi people - I am late to the party, this is my first open source PR :). Let me know if I need to make any changes.

- Changes to group issues by categories on the 'more issues' screen. No changes to Home: still shows just the 'active' issues.
- Issues are now fetched using the all==true flag — that means issues are not refetched again on the 'more issues' screen.

Details:
- Fetching issues:
	- Removed the Query (active, inactive) enum - now FetchIssuesOperation fetches
	all issues whenever 'fetchIssues' method IssuesManager is called.
	- IssuesViewController now does not create its own instance of IssuesManager.
	Instead the caller has to create and pass IssuesManager instance to IssuesViewController.
	This way issues that are already downloaded are shared between Home and More issues
	screen (this is because Home passes its own IssueManager reference to More issues in 
	prepare segue).
	
- Showing only 'active' issues on Home:
	- Because IssuesManager now fetches all issues - IssuesViewController has to selectively
	display only 'active' issues.
	- Instead of filtering on the main thread - hiding the row (by returning height 0) for 
	inactive issues.

- Grouping issues by category on more issues screen:
	- After downloading all issues, IssuesManager creates the category->[Issues] relationship.
	- This step also involves creating 'Uncategorised issues' section for issues that are not part
	of any category (this matches with the way things are implemented on the web front end)
	- Finally categories are alphabetically sorted. All three steps happen on a BG thread.

- UI changes:
- Home:
	- 'Whats important' header is now not sticky (scrolls away as the user scrolls). This is because earlier this
	was implemented as a section header - whereas in this PR 'whats important' is now a tableHeader.
	- This impacts Home.

- More issues screen:
	- 'Whats important' header is now NOT displayed on the more issues screen - I am not sure if this is expected
	behaviour. Did this because this header along with the first category header was not looking nice.
	- Category headers are displayed in the same style as existing header style - may be we can improve the UI a bit here (see the 
	attached screen shots).
	- Uncategorised issues are always shown last - after all other categories.


indraShan:
Attaching screenshots. 
uncategorised
home
more_issues

nickoneill:
This looks great, thank you! I'll send you back a review later today.

Abizern:
It does look good. I'm not at a computer right now but I see you've changed the Issue model. I think this may break UITests. Could you try it, please? My apologies if this doesn't.

nickoneill:
Thanks for consolidating the request to `all` rather than making two requests, that makes things a bit simpler.

A few notes/questions:
* Hiding the inactive issues with height=0: isn't it easier to filter the list to do active=true rather than hack around not showing inactive cells?
* Not sure I understand why "What's important" isn't a stick header on the main screen anymore - this could even be mocked as a fake category if you wanted to reuse the same code.
* "What's important" on More Issues: it's OK to leave this out now that we have the category headers
* @Abizern is correct, the snapshot test fails, probably because of the model change

subdigital:
Agree on avoiding `height=0` cells. It's better to filter these out in our datasource.

indraShan:
Hiding rows for inactive issues:
- This was easiest way I could think of - but yes does not feel like the 'correct' way of doing it.
- Another approach could be that the category object could have two properties - activeIssues and inactiveIssues. These two properties would be set right after the issues are downloaded (in IssuesManager) - this would happen on BG thread. IssuesViewController's viewModel will choose the correct property based on whether it is showing all or only active issues. What do you guys think? I will try to implement this, if you guys don't see any issues with this approach.

Whats important header on Home not sticky anymore: 
- You are right - will make it a section header again on Home. Don't know why didn't do it in the first place :)

Failing test case:
- Will checkout the failing test. Should have ran the tests before pushing the PR.

nickoneill:
Issues already have an `active` property, I'm not seeing a reason to keep track of this elsewhere.

You can get a list of the active issues from the full list with `let activeIssues = allIssues.filter({ $0.active == true })` in one step.

The other thing that's bothering me is the dual use of the new category model. It's both the child property of issue (i.e. deserialized with the issues) and also being used as the parent of an array of issues for the more issues page. It seems like the model approach is correct for deserialization, but when building the category list on "more issues", it might make more sense as `[[String: [Issue]]]` (i.e. an array of maps that connect a string to a list of issues), or a separate struct of similar construction of that feels too complex.

Could use your thoughts here ^ @subdigital when you've got a minute.

indraShan:
@nickoneill We need filtered list in numberOfRows in section and again in cellForRowAtIndexPath. I don't think filtering on the fly is a good idea.

Agree that issue to category mapping is not required. Had kept it just in case we need to get the category for a particular issue - but not being used anywhere. Right now we only need to know issues in a category. 

Had thought of using a dictionary at first - Category as the key and the Issues that belong to that category as value. But we want the categories (i.e. keys) alphabetically sorted on the more issues screen - which adds another sorting step if we use a dictionary.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -55,11 +61,17 @@ class IssuesViewController : UITableViewController {
...
+        // Refetch data only if we don't have anything to display.
+        if viewModel.categories.count == 0 {
+            loadIssues()
+        }
+
+        // 'Whats important' header gets shown only when we are showing
+        // active issues only.
+        if !shouldShowAllIssues {
+            tableView.tableHeaderView = headerWithTitle(title: R.string.localizable.whatsImportantTitle())
+        }
```
subdigital: Thinking this would be simpler to show as a section header, so we wouldn't need this little bit of complexity...

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -118,13 +128,26 @@ class IssuesViewController : UITableViewController {
         lastLoadResult = result
         issuesDelegate?.didFinishLoadingIssues()
         if case .success = result {
-            viewModel = ViewModel(issues: issuesManager.issues)
+            viewModel = ViewModel(categories: issuesManager.categories)
```
subdigital: Here we do have _all_ the issues, but we're handing over a specific view of those issues by grouping them into categories. What if we moved this categorization/group _into_ the view model, then we'd still have access to the entire array of issues for the Active Issues mode, and the grouped for the All Issues list.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -152,70 +175,93 @@ class IssuesViewController : UITableViewController {
...
+            typedInfo.destination.issue = viewModel.categories[indexPath.section].issues[indexPath.row]
         }
     }
     
     // MARK: - UITableViewDataSource
     
     override func numberOfSections(in tableView: UITableView) -> Int {
-        return 1
+        // Empty tableview if we haven't fetched data yet.
+        if viewModel.categories.count == 0 {
```
subdigital: this is circumstantial code. I'd prefer do have an `isLoading` property or something similar that isn't tied to an arbitrary dictionary or array having data.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -152,70 +175,93 @@ class IssuesViewController : UITableViewController {
...
 
+    override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        guard indexPath.row < viewModel.categories[indexPath.section].issues.count else {
+            return UITableViewAutomaticDimension
+        }
+        let issue = viewModel.categories[indexPath.section].issues[indexPath.row]
+        // Hide the row (height 0) if the issue for that cell is inactive and
+        // we are currently showing active issues only.
+        if !shouldShowAllIssues && issue.inactive {
+            return 0
```
subdigital: as discussed before, I really want to avoid doing this. Lean on the model to return the correct number of cells rather than faking it by making them zero height.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -152,70 +175,93 @@ class IssuesViewController : UITableViewController {
...
+        }
+        return UITableViewAutomaticDimension
+    }
+
+    override func tableView(_ tableView: UITableView, estimatedHeightForRowAt indexPath: IndexPath) -> CGFloat {
+        return 75
+    }
```
subdigital: This can be set in the storyboard (or in `viewDidLoad`) since it's just a static value and doesn't vary by section, etc...

FiveCalls/FiveCalls/IssuesViewController.swift
indraShan: I don't think I understand you.

In this PR - the table view always has as many section as the number of unique categories. Its just that when we want to show active issues - section headers are not shown. So if there are 7 categories in the response - the number of sections displayed by the tableView would be 7. But in the viewForHeaderInSection we return 'nil' because we don't want to group issues by category. As a result Issues appear as if they are in a single section.

So currently we are handing over the entire <i>view</i> to the view model - not a subset of it. Also each Category object (inside the categories array) contains all the issues that belong to it - active and inactive both. If a specific Issue inside a category is inactive and if we are currently not showing all issues - the cell is hidden. 

Also if we pass the response (all issues) as is to the ViewModel and let it process it (sort, associate categories), that work would be done on the main thread - which is not ideal (again I am not sure if I understand your original comment, so this last paragraph could be illogical)

FiveCalls/FiveCalls/IssuesViewController.swift
indraShan: Will do this change. Thanks.

FiveCalls/FiveCalls/IssuesViewController.swift
indraShan: Got it. Will change this.

FiveCalls/FiveCalls/IssuesViewController.swift
indraShan: I am not sure if estimated height can be set from the Storyboard. I know height can be, but not sure about estimated height. Will take a look. Otherwise will just set this from viewDidLoad. Thank you.

FiveCalls/FiveCalls/IssuesViewController.swift
indraShan: I agree. But I haven't being able to figure out a way to do this, without using the main thread to do the data processing (except the current implementation and the alternate implementation I suggested in an earlier comment ie. two properties on the viewModel - activeIssues and inactiveIssues).

The requirements are:
- Show active issues on Home - don't care to which categories they belong.
- Show all issues on More - group them in categories, with categories alphabetically sorted.

ViewModel has to basically vend data for both these scenarios - so we could move some of the logic from controller to ViewModel and get rid of height==0 logic if we pre-process the data on a BG thread before handing it over to ViewModel. That way viewModel will return the correct number of cells (because it knows exactly how many activeIssues exist for a category) - removing the need for height=0 zero hack.

FiveCalls/FiveCalls/IssuesViewController.swift
subdigital: Estimated height can definitely be set in the storyboard.

indraShan:
Figured out why the tests were failing - had to update the fake responses to include the category and active status. 

To ship this PR now we need to figure out the right fix to show active issues...

FiveCalls/FiveCalls/IssuesViewController.swift
subdigital: This is essentially what I'm suggesting.  Take the raw list, create whatever representations you need of this data in the view model (perhaps in the background, or even one once, it's not a huge deal if on the main thread), and then just use UITableView how it was intended.

In the long run hacks like these will end up hurting us more than helping, so I think this approach will serve us better in the end.

indraShan:
Hi Guys,  
I have updated the code to remove the cell hide hack and fix the failing test case. Also 'whats important' header on Home is now sticky (which matches old behaviour). Moved the ViewModel to its own file - IssuesViewModel. 
Let me know what you think. Thank you.

FiveCalls/FiveCalls/IssuesViewModel.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    func numberOfSections() -> Int {
+        if showAll {
+            return categories.count
+        }
+        return 1
+    }
+
+    func numberOfRowsInSection(section: Int) -> Int {
+        if showAll {
```
subdigital: This entire class splits functionality based on this flag. This is calling to be a different type :)

At this point I think it makes sense to have 2 view models:

- `ActiveIssuesViewModel`  and `CategorizedIssuesViewModel` (not married to that name)

Then we can push the appropriate logic into the right view model type.

subdigital: We'll also need to extract out a protocol for these 2 so the view controller can treat them equally.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -52,7 +49,7 @@ class IssuesViewController : UITableViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         // Construct viewModel from the issues already fetched by the caller.
-        viewModel = ViewModel(categories: self.issuesManager.categories)
+        viewModel = IssuesViewModel(categories: self.issuesManager.categories, showAll: shouldShowAllIssues)
```
subdigital: This is where we would build our view model instance.  Something like:

```swift
viewModel = shouldShowAllIssues ? CategorizedIssuesViewModel(issues: self.issuesManager.issues) : ActiveIssuesViewModel(issues: self.issuesManager.issues)
```

Since we're splitting this logic at the view model, it probably makes sense to push the categorization of issues back up into `IssuesManager` so we can hand the appropriate collection to the view model.

FiveCalls/FiveCalls/IssuesViewModel.swift
indraShan: Makes sense. This will make the code much more cleaner - will try to get this in by today evening.

indraShan:
Changes to fix last set of review comments. Now we have a new protocol and two view models that conform to that protocol. AllIssuesViewModel shows all categories grouped by categories whereas ActiveIssuesViewModel shows only the active ones in a single section. Had missed a change in my check-in because of which the app would crash on iPad after tapping a issue row - now fixed.

indraShan:
Hi guys, what is the next step for this PR? Let me know if any further changes are required.

PR #195, opened 2017-10-16
Remote notification requesting

nickoneill:
* will request remote notification permissions after a user completes their first set of calls
* currently no UI for turning these off once they're on (other than the system UI)


FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -39,9 +40,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        oneSignalStartup(launchOptions: launchOptions)
+
         return true
     }
 
+    func oneSignalStartup(launchOptions: [UIApplicationLaunchOptionsKey: Any]?) {
+        let onesignalInitSettings = [kOSSettingsKeyAutoPrompt: false]
+
+        OneSignal.initWithLaunchOptions(launchOptions,
+                                        appId: "5fd4ca41-9f6c-4149-a312-ae3e71b35c0e",
```
subdigital: Would be nice to tuck this inside a plist so we could change it by a configuration change (by scheme perhaps).  Then we could access it with `AppConfiguration.current.setting(for: "OneSignalAppId")`.  What do you think?

FiveCalls/FiveCalls/CallScriptViewController.swift
```diff
@@ -189,7 +190,27 @@ class CallScriptViewController : UIViewController, IssueShareable {
         present(alertController, animated: true, completion: nil)
     }
 
-    
+    func checkForNotifications() {
+        let permissions = OneSignal.getPermissionSubscriptionState()
+
+        if permissions?.permissionStatus.hasPrompted == .some(false) {
```
subdigital: Don't need the `.some(...)` here, `false` will work.

FiveCalls/FiveCalls/CallScriptViewController.swift
```diff
@@ -189,7 +190,27 @@ class CallScriptViewController : UIViewController, IssueShareable {
...
+            let alert = UIAlertController(title: R.string.localizable.notificationTitle(), message: R.string.localizable.notificationAsk(), preferredStyle: .alert)
```
subdigital: `.notificationAsk` made me chuckle. [;)](http://askisnotanoun.com)

FiveCalls/FiveCalls/CallScriptViewController.swift
```diff
@@ -189,7 +190,27 @@ class CallScriptViewController : UIViewController, IssueShareable {
...
+            alert.addAction(UIAlertAction(title: R.string.localizable.notificationAll(), style: .default, handler: { (action) in
+                OneSignal.promptForPushNotifications(userResponse: { (success) in
+                    OneSignal.sendTag("all", value: "true")
+                })
+            }))
+            alert.addAction(UIAlertAction(title: R.string.localizable.notificationImportant(), style: .default, handler: { (action) in
+                OneSignal.promptForPushNotifications(userResponse: { (success) in
+                    //
```
subdigital: Are you going to tag with `"important"` here?  What's the thinking?

FiveCalls/NotificationsService/NotificationService.swift
```diff
@@ -0,0 +1,36 @@
+//
```
subdigital: we have 2 of these?

FiveCalls/NotificationsService/NotificationService.swift
```diff
@@ -0,0 +1,36 @@
...
+    var contentHandler: ((UNNotificationContent) -> Void)?
+    var bestAttemptContent: UNMutableNotificationContent?
+
+    override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+        self.contentHandler = contentHandler
+        bestAttemptContent = (request.content.mutableCopy() as? UNMutableNotificationContent)
+        
+        if let bestAttemptContent = bestAttemptContent {
+            // Modify the notification content here...
+            bestAttemptContent.title = "\(bestAttemptContent.title) [modified]"
```
subdigital: This is just example code? Probably should remove.

FiveCalls/Podfile
```diff
@@ -10,12 +10,17 @@ end
 
 target 'FiveCalls' do
   common_pods
+  pod 'OneSignal', '>= 2.5.2', '< 3.0'
```
subdigital: `pod 'OneSignal', ~> '2.5'` is more what we want. Meaning, _give me the highest in the 2.x range_.

FiveCalls/Podfile
```diff
@@ -10,12 +10,17 @@ end
...
   pod 'CPDAcknowledgements'
   pod 'Fabric'
   pod 'Crashlytics', '~> 3.8.4'
   pod 'DZNEmptyDataSet', git: 'https://github.com/subdigital/DZNEmptyDataSet'
 end
 
+target 'NotificationsService' do
+    pod 'OneSignal', '>= 2.5.2', '< 3.0'
```
subdigital: same here

FiveCalls/FiveCalls/CallScriptViewController.swift
nickoneill: Having not used one signal segments / tags before, I was going to use no tags as an indication of only important messages, and the "all" tag for all messages. There's no level of push less than "just important messages" so it feels like a good baseline to me.

FiveCalls/NotificationsService/NotificationService.swift
nickoneill: Oh yeah, for sure. I was thinking this was all the boilerplate needed without even looking at it.

FiveCalls/FiveCalls/CallScriptViewController.swift
nickoneill: 👍

FiveCalls/Podfile
nickoneill: 👍

FiveCalls/Podfile
nickoneill: 👍

FiveCalls/NotificationsService/NotificationService.swift
nickoneill: Weird. Seems like the one signal docs suggest modifying the one in the notification extension, I'll keep that one.

FiveCalls/FiveCalls/AppDelegate.swift
nickoneill: Put this in build settings, then referenced it in Info.plist so it can be accessed through `Bundle.main.infoDictionary`

nickoneill:
Revised from feedback. We can wait on taking you directly to the issue.

nickoneill:
Rebased on changes from @mccarron, ping for review @subdigital

PR #197, opened 2017-10-31
Fix iPhone X header positioning issue

bjtitus:
Fixes some minor issues with positioning on iPhone X with the notch.

napkin 10-30-17 10 10 48 pm

A simple change to respect the top layout guide for the header, and update content inset whenever subviews change frame sizes.


nickoneill:
Never had before and after shots before that have such nice ligatures, looks great 😉

bjtitus:
I totally missed that bug! @nickoneill One of my favorite uses of Napkin: http://aged-and-distilled.com

Issue #180, opened 2017-10-06
Logo in navigation bar doesn't play well with the iPhone X notch


Abizern:
Still a couple of weeks left to solve this.

![simulator screen shot - iphone x - 2017-10-07 at 00 09 08](https://user-images.githubusercontent.com/6902/31301792-96bd5514-aaf4-11e7-9cb2-d8ead37d7e41.png)

Abizern:
Who's the lead on design for the mobile apps?

subdigital:
It's been a group effort.  I started the app following the design of the web app. Later started to move more toward more native conventions, after receiving some excellent design feedback from , which has been addressed somewhat, but not entirely.

@ianyh @nickoneill and myself can provide some direction or validation if you have a specific question, otherwise just go for what you think will work well and looks like it belongs on the platform.

nickoneill:
Playing around with some options that would shorten the header for all platforms...

<img width="771" alt="screen shot 2017-10-09 at 5 53 07 pm" src="https://user-images.githubusercontent.com/49688/31364641-067e2eb4-ad1b-11e7-9639-40deda05b549.png">
<img width="757" alt="screen shot 2017-10-09 at 5 53 41 pm" src="https://user-images.githubusercontent.com/49688/31364642-0696436e-ad1b-11e7-9138-b610d8a0d681.png">
<img width="757" alt="screen shot 2017-10-09 at 5 54 21 pm" src="https://user-images.githubusercontent.com/49688/31364643-06af24a6-ad1b-11e7-8aa3-f00e716743dc.png">

Definitely liking the idea of ditching the full logo text for the stars alone. I'm sure some executive somewhere is rolling in their grave "they didn't make the logo bigger..."

Abizern:
FWIW, I like the second header. I think it looks clean and the notification button is at the top right which is more natural for most people.

Maybe to keep to the colour scheme, you could make the top star of the logo red?

nickoneill:
Yeah, the second option feels good too. We've used the stars alone before, I think we can get away with it here without changing the colors.

If notifications here is going to be both local reminders and options for remote, I wonder if it makes more sense to make it a "settings" gear and then drop you into the about controller where notification settings can live alongside about info.

Abizern:
Using a gear might be a better option. Looking at this again, having a rendered bell rather than a line icon (if you know what I mean) seems a bit out of kilter with modern UI.

subdigital:
[...]

PR #200, opened 2017-12-28
Top Bar Redesign

mccarron:
This addresses the new design idea in #180 for the new top bar. Adjusted the layout constraints to handle iPhone pre-X in both landscape and portrait, iPads and the X as well. See screenshots attached. 

Debated making the 5 Stars logo a button, maybe open the About screen, so I kept it a button but it's disabled for now. We could make it open About and remove the bottom button. Maybe an idea for a future enhancement? 

simulator screen shot - iphone 5s - 2017-12-28 at 14 44 15
simulator screen shot - iphone x - 2017-12-28 at 14 37 11
simulator screen shot - iphone x - 2017-12-28 at 14 37 17
simulator screen shot - iphone 5s - 2017-12-28 at 14 44 16
simulator screen shot - ipad air 2 - 2017-12-28 at 14 46 01
simulator screen shot - ipad air 2 - 2017-12-28 at 14 46 02


subdigital:
Thanks for getting this done.  Agree about it looking like a button. If we went the route of making that open about and we removed the button at the bottom, then we might as well just move the logo to a tab bar item and call that About. 

For now I think this addresses the immediate concerns.

mccarron:
Yeah I agree, a tabbed interface may be better path forward for the bottom.

Issue #199, opened 2017-12-20
auth0 Login


nickoneill:
Let's implement login so we can doing things like track stats across devices.

Lacking design at the moment, so let's keep things simple and we can improve as the functionality comes together. Spec and auth0 keys are on slack.

nickoneill:
Fixed

PR #201, opened 2018-01-15
User Logins with Auth0

melville:
This PR addresses issue #199 and adds a login capability to the app via Auth0 and includes the user's ID in call reports if they're logged in.  The user can log in or out in a new section at the top of `MyImpactViewController`:

simulator screen shot - iphone 8 - 2018-01-15 at 12 24 46

simulator screen shot - iphone 8 - 2018-01-15 at 12 24 33


subdigital:
Trying this out now...

subdigital:
image

Is it possible to get a higher res image?  The one being used here looks really low res.

Also, I think for Twitter or Facebook login, we should use the `screen_name` for the subtitle, since we don't have an email.

(I think this is called `username` for facebook, but I don't remember exactly)

FiveCalls/FiveCalls/MyImpactViewController.swift
```diff
@@ -58,7 +66,27 @@ class MyImpactViewController : UITableViewController {
...
+        signInContainer.layer.shadowColor = UIColor.black.cgColor
+        signInContainer.layer.shadowOffset = CGSize(width: 1.0, height: 1.0)
+        signInContainer.layer.shadowOpacity = 0.2
+        signInContainer.layer.shadowRadius = 1.2
+
+        profileContainer.layer.masksToBounds = false
+        profileContainer.layer.shadowColor = UIColor.black.cgColor
+        profileContainer.layer.shadowOffset = CGSize(width: 1.0, height: 1.0)
+        profileContainer.layer.shadowOpacity = 0.2
+        profileContainer.layer.shadowRadius = 1.2
```
subdigital: This shadow setup is identical. I'd favor moving all of this into a UIView subclass.

Something like `ProfileCard`? We can move all this common code there (as well as the runtime corner radius attribute in the Storyboard.

Then we can update both IBOutlets to be of that class:

image

subdigital: Also, for the shadow offset: `CGSize(width:0, height: 1)`  (light source from top)

FiveCalls/Podfile
```diff
@@ -10,6 +10,8 @@ end
 
 target 'FiveCalls' do
   common_pods
+  pod 'Auth0', '~> 1.9'
+  pod 'Kingfisher', '~> 4.0'
```
subdigital: Now that we have `Kingfisher` (which is 👌 ), can you remove `RemoteImageView` and set those images with Kingfisher instead?  There is a race condition with my RemoteImageView that I never bothered fixing, which causes the wrong image to sometimes be displayed on contact rows.

FiveCalls/FiveCalls/MyImpactViewController.swift
```diff
@@ -69,13 +97,57 @@ class MyImpactViewController : UITableViewController {
...
+        tableView.tableHeaderView = headerView
     }
     
-    override func viewWillLayoutSubviews() {
-        super.viewWillLayoutSubviews()
-        if let headerView = tableView.tableHeaderView {
-            let height = subheadLabel.isHidden ? impactLabel.frame.maxY : (subheadLabel.frame.maxY + 25.0)
-            headerView.frame.size.height = height
+    @IBAction func loginButtonTapped(_ sender: Any) {
+        if (SessionManager.shared.userIsLoggedIn()) {
```
mccarron: Unneeded `()` around the if condition.

FiveCalls/FiveCalls/IssuesContainerViewController.swift
```diff
@@ -73,8 +73,12 @@ class IssuesContainerViewController : UIViewController, EditLocationViewControll
         configureChildViewController()
         setupHeaderWithBlurEffect()
         editRemindersButton.tintColor = .fvc_darkBlue
+        locationButton.tintColor = .fvc_darkBlue
         let image = UIImage(named: "gear")?.withRenderingMode(.alwaysTemplate)
         editRemindersButton.setImage(image, for: .normal)
+        if (SessionManager.shared.userHasCredentials()) {
```
mccarron: Unneeded `()` around the if condition.

FiveCalls/FiveCalls/MyImpactViewController.swift
```diff
@@ -58,7 +66,27 @@ class MyImpactViewController : UITableViewController {
         streakLabel.text = weeklyStreakMessage(for: weeklyStreakCount)
         let numberOfCalls = viewModel.numberOfCalls
         impactLabel.text = impactMessage(for: numberOfCalls)
-        subheadLabel.isHidden = numberOfCalls == 0
+        if (numberOfCalls == 0) {
```
mccarron: Unneeded `()` around the if condition.

subdigital:
This looks great. Thanks for putting this together!

melville:
Thanks for the merge @subdigital!  To answer your questions:

* I don't think we can get  higher-res image unless we manage user images ourselves.  We're only provided a single URL by Auth0.
* For the differences between the different auth providers, I haven't actually tried comparing values from the various supported auth services, but it seems like Auth0 is managing the differences under its hood.  This is the list of values that are available to us in the user profile:

```
    public let sub: String    // This is the unique ID

    public let name: String?
    public let givenName: String?
    public let familyName: String?
    public let middleName: String?
    public let nickname: String?
    public let preferredUsername: String?

    public let profile: URL?
    public let picture: URL?
    public let website: URL?

    public let email: String?
    public let emailVerified: Bool?

    public let gender: String?
    public let birthdate: String?

    public let zoneinfo: TimeZone?
    public let locale: Locale?

    public let phoneNumber: String?
    public let phoneNumberVerified: Bool?

    public let address: [String: String]?
    public let updatedAt: Date?
```

For the subtitle I'm checking for an email address, and if that's not available I use the 'name' field, which seems to be populated with the email address by default if you're a user that authenticates via email.  I'm not sure what the best policy is moving forward, but it should be relatively easy to tweak the profile to use different variables if we discover those to be unsuitable.

PR #209, opened 2018-04-11
Shift user stats to the server

melville:
This PR takes some steps towards moving user stats to the server.  We've already introduced code that ties users to call reports, but now we will actually fetch the server-side stats when reporting on the user's impact.  We will also send the user's unreported local stats any time they start a session by logging in.  This means that whoever logs in first on a device with existing stats will get credit for those unreported calls.  I think this is fine since it seems very unlikely that a different user than the one who created those stats would log in, but if you have concerns let me know.

Also note that this code does not purge the local stats after reporting them, instead it flags them as reported so that they won't be sent twice.  It would be simpler to purge them after reporting, but currently we don't have another mechanism for determining whether a user has completed calls for a particular issue so that we can display the status appropriately in the issues list.  In the future we could make the API call for fetching issues aware of the user that's requesting the list so that it could include completion status, and that would eliminate the need to store any stats on the device (beyond those which haven't been reported yet).

I have done some light testing on these changes and they appear to work properly, but it would be good to do more thorough testing before pushing a new version to iTunes.  It would also be nice to change the UI so that we show spinners for the stats while they're still loading.  Currently we just display the local unreported stats until the call to fetch user stats from the server is finished.


FiveCalls/FiveCalls/ContactLog.swift
```diff
@@ -42,21 +44,23 @@ extension ContactLog : Storable {
...
     }
     
     static func ==(lhs: ContactLog, rhs: ContactLog) -> Bool {
         return lhs.issueId == rhs.issueId &&
                 lhs.contactId == rhs.contactId &&
                 lhs.phone == rhs.phone &&
-                lhs.outcome == rhs.outcome
+                lhs.outcome == rhs.outcome &&
+                lhs.reported == rhs.reported
     }
```
subdigital: I think we can remove these `==` and `hashValue` methods right?  In Swift 4.1 these should automatically use all members if the object is `Equatable`/`Hashable`.

FiveCalls/FiveCalls/FetchUserStatsOperation.swift
```diff
@@ -0,0 +1,79 @@
...
+    var userStats: UserStats?
+    var httpResponse: HTTPURLResponse?
+    var error: Error?
+    
+    override func execute() {
+        
+        // We'll need the user's access token to authenticate the request
+        SessionManager.shared.credentialsManager.credentials {
+            guard $0 == nil else {
+                self.error = $0
```
subdigital: I think we should avoid using implicit parameters like `$0`, `$1` for this method, since it's quite long and give them names for clarity.

FiveCalls/FiveCalls/FetchUserStatsOperation.swift
```diff
@@ -0,0 +1,79 @@
...
+                self.finish()
+                return
+            }
+
+            let credentials = $1
+            if let idToken = credentials?.idToken {
+        
+                let config = URLSessionConfiguration.default
```
subdigital: This block is ripe for extract method:  maybe `fetchStats(withToken:)` ?

FiveCalls/FiveCalls/ImpactViewModel.swift
```diff
@@ -9,30 +9,35 @@
 import Foundation
 
 struct ImpactViewModel {
-    let logs: [ContactLog]
     
-    init(logs: [ContactLog]) {
+    let userStats: UserStats?
+    let logs: ContactLogs
+    
+    init(logs: ContactLogs, stats: UserStats?) {
```
subdigital: Just so I have this clear, this is a transitional thing...  we need to keep around ContactLogs for other areas of the app currently, but until we unify the entire thing we have to merge these 2 sets?

FiveCalls/FiveCalls/MyImpactViewController.swift
```diff
@@ -61,31 +62,22 @@ class MyImpactViewController : UITableViewController {
...
+        totalCallsOp.completionBlock = {
+            self.totalCalls = totalCallsOp.numberOfCalls
             DispatchQueue.main.async {
                 self.tableView.reloadData()    
             }
-            
         }
-        OperationQueue.main.addOperation(op)
-        
+        OperationQueue.main.addOperation(totalCallsOp)
```
subdigital: Why main queue?

FiveCalls/FiveCalls/MyImpactViewController.swift
```diff
@@ -117,6 +122,22 @@ class MyImpactViewController : UITableViewController {
...
+        if SessionManager.shared.userIsLoggedIn() {
+            // Fetch the user's call stats
+            let userStatsOp = FetchUserStatsOperation()
+            userStatsOp.completionBlock = {
+                self.userStats = userStatsOp.userStats
+                DispatchQueue.main.async {
+                    self.updateStats()
+                }
+            }
+            OperationQueue.main.addOperation(userStatsOp)
```
subdigital: same here, I think we should use a background queue...

FiveCalls/FiveCalls/ReportUserStatsOperation.swift
```diff
@@ -0,0 +1,79 @@
...
+        
+        // We'll need the user's access token to authenticate the request
+        SessionManager.shared.credentialsManager.credentials {
+            guard $0 == nil else {
+                self.error = $0
+                self.finish()
+                return
+            }
+            
+            let credentials = $1
```
subdigital: Same comment about unnamed parameters...

FiveCalls/FiveCalls/ReportUserStatsOperation.swift
```diff
@@ -0,0 +1,79 @@
...
+                let task = session.dataTask(with: request) { (data, response, error) in
+                    if let e = error {
+                        self.error = e
+                    } else {
+                        self.httpResponse = response as? HTTPURLResponse
+                        if let status = self.httpResponse?.statusCode, status >= 200, status <= 299 {
+                            // Mark all unreported stats as reported, now that we've submitted them all
+                            unreportedLogs.forEach() { log in
+                                self.logs.markReported(log)
+                            }
```
subdigital: Think its worth making the method batch this to avoid saving repeatedly?

FiveCalls/FiveCalls/SessionManager.swift
```diff
@@ -77,11 +103,44 @@ class SessionManager {
...
+        return Promise { seal in
+            if userIsLoggedIn() {
+                let logs = ContactLogs.load()
+                let unreportedLogs = logs.unreported()
+                if unreportedLogs.count > 0 {
+                    // Send all of our unreported stats to the server.
+                    let reportStatsOperation = ReportUserStatsOperation(logs: logs)
+                    reportStatsOperation.completionBlock = {
+                        if let status = reportStatsOperation.httpResponse?.statusCode {
+                            if status >= 200 && status <= 299 {
```
subdigital: We could leverage pattern matching here `if case (200..<300) = status {`

FiveCalls/FiveCalls/ImpactViewModel.swift
melville: `ContactLogs` wouldn't be strictly necessary if the server were providing completion status for issues, but it would be useful to keep it so that we can keep track of calls that we weren't able to report to the server due to connectivity issues or server downtime.  In that scenario we would only keep the `ContactLogs` for unreported calls, instead of keeping them all on the device forever the way we do now.  If we got completion data from the server we could wipe the local entry as soon as we were able to report it.  As of now though we definitely still need to keep all of the local stats so we can use them to mark up the issues list properly to show what the user has already done.

FiveCalls/FiveCalls/MyImpactViewController.swift
melville: This wasn't a conscious choice on my part, I was just following the established logic used by the existing Operations.  It looks like we do mark our Operations as asynchronous, and their logic operates via callbacks, so I'm not sure it matters very much, but we could certainly create another queue for this if you want.

Issue #214, opened 2018-06-22
Markdown not parsing in descriptions and scripts


nickoneill:
I just tested a couple markdown examples in the description and script and it doesn't look like they're parsing correctly to styled text.

Can you take a look @mccarron? Let me know if you need me to add test markdown to an older issue, I removed it from this issue after I tested.

Both of the screens below have text that has `__test markdown__` which should be bold

nickoneill:
![img_3174](https://user-images.githubusercontent.com/49688/41789305-b2eb6700-7603-11e8-8fdb-802273078ec4.PNG)
![img_3175](https://user-images.githubusercontent.com/49688/41789306-b32a4e5c-7603-11e8-932b-9327ae099671.PNG)

mccarron:
Yeah we didn't add Markdown support yet just a pod that adds the ability to parse Markdown. You said the API didn't have Markdown yet so there wasn't anything to do. ;) 

Can we get some Markdown onto a Staging/Test API to build this out?

nickoneill:
I don't have a staging API for the main issues list (it's getting moved to a different server soon that already has a staging API) but I added some test markdown to an issue on the more issues list: "Protect the Right to Fight Workplace Sexual Harassment"

Let me know if that works OK.

mccarron:
Sure that'll work, I'll take a look at hooking this up either tomorrow or Sunday then.

nickoneill:
🏓

nickoneill:
Done!

PR #215, opened 2018-06-27
Markdown Rendering Added

mccarron:
First pass at fixing #214. I am using the CSS file used for other parts of the app for now (see results below). 

Should we create a custom CSS just for Issues? 

simulator screen shot - iphone x - 2018-06-26 at 20 09 02

Also the app takes a lot longer to startup due to Markdown parsing. May look to what we can do to speed that up a bit. May compare other Markdown parsers for speed.

### DIFF
--- a/FiveCalls/FiveCalls/Issue.swift
+++ b/FiveCalls/FiveCalls/Issue.swift
@@ -7,33 +7,25 @@
 //
 
 import Foundation
-
-/*
- 
- {
- "contacts": [],
- "id": "recNqspdW0JHxisu8",
- "name": "Reject the Appointment of Steve Bannon to the National Security Council",
- "reason": "Trump has removed the existing Director of National Intelligence and Chairman of Joint Chiefs of Staff from the \u201cprincipals committee\u201d of the National Security Council and replaced them with his Chief Strategist, Steve Bannon. Bannon is the former leader of Breitbart News, an anti-Semitic and white supremacist right-wing opinion and \"news\" outlet whose notable output includes articles comparing Planned Parenthood to the Holocaust. \n\nThe purpose of the National Security Council is to ensure that the president has the best possible advice from his Cabinet, the military, and the intelligence community before making major decisions.  Steve Bannon not only lacks the considerable experience needed to be the principal foreign policy adviser to the president, he has personally avowed a desire to \"bring everything crashing down, and destroy all of today\u2019s establishment.\" Allowing Trump to continue to be guided by Bannon in matters of national security will bring about the deaths of many innocent people. ",
- "script": "Hi, my name is [NAME] and I\u2019m a constituent from [CITY, ZIP]. \n\nI\u2019m calling today becuase I am outraged by the appointment of Steve Bannon to the National Security Council. I expect [Senator/Rep's Name]  to take an immediate stand against both Bannon's appointment and the removal of more experienced staff from the \"principals committee\".                           \n\nThank you for your hard work answering the phones.\n\n[IF LEAVING A VOICEMAIL: please leave your full street address to ensure your call is tallied] ",
- "categories": [{
-   "name": "Foreign Affairs"
- }],
- "inactive": false,
- "outcomes": ["unavailable", "voicemail", "contact", "skip"]
- }
- 
- */
+import Down
+import Rswift
 
 struct Issue {
     let id: String
     let name: String
-    let reason: String
+    let reason: NSAttributedString
     let script: String
     let category: Category?
     let inactive: Bool
     let contacts: [Contact]
     let outcomes: [Outcome]
+
+    static var style: String {        
+        if let bytes = try? Data(resource: R.file.stylesCss)  {
+            return String(decoding: bytes, as: UTF8.self)
+        }
+        return ""
+    }
 }
 
 extension Issue : JSONSerializable {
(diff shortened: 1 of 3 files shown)


nickoneill:
We're moving away from the blue text in general (and using Roboto in the app), so I'd rather have it look like it did before the markdown renderer. Looks like some excessive side padding too - we should probably start the CSS fresh for this section.

App startup is longer, or generating the issue detail controller is longer?

FiveCalls/FiveCalls/Issue.swift
```diff
@@ -7,33 +7,25 @@
...
- */
+import Down
+import Rswift
 
 struct Issue {
     let id: String
     let name: String
-    let reason: String
+    let reason: NSAttributedString
     let script: String
```
nickoneill: Script should also be markdown-compatible

subdigital:
I think we should move to a more standard body font color. Since blue is used for the tintColor for interactive elements I think this is probably the wrong choice for body text.

subdigital:
Derp. What Nick said...

mccarron:
Correct I said the CSS was just reused from elsewhere and only temporary so we can make one. 

I just cloned it to it's own file so @nickoneill can hack on the CSS.

mccarron:
App starts up on time just takes longer to parse all the data, just see a spinner. Maybe at next version we can add a better "loading" screen and optimize the parsing a bit.

subdigital:
The margins on this css file don't really fit here, @nickoneill are you working on tweaking the CSS file?

nickoneill:
I am, will have something else later tonight.

nickoneill:
Wow, I thought my network was being slow but that's actually the markdown rendering time!

nickoneill:
A few things while hacking on this:

* I think we have to change the UILabel to `UITextView` and implement a delegate method to get links working from within `NSAttributedString`
* I swapped the markdown rendering phase to the view detail loading so we don't process every markdown regardless of showing it - there's a slight lag when loading the controller but it's only a single issue text now. 

I'll continue to work on this tomorrow a bit.

FiveCalls/FiveCalls/Info.plist
```diff
@@ -19,22 +19,22 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.4</string>
+	<string>1.5</string>
```
subdigital: Don't forget that you also have to bump the extension version when you do this.

FiveCalls/FiveCalls/Issue.swift
```diff
@@ -55,6 +46,18 @@ extension Issue : JSONSerializable {
...
+//        if let converted = try? markdown.toAttributedString(.default, stylesheet: Issue.style) {
+//            attributedReason = converted
+//        }
+
+        var attributedScript = NSAttributedString(string: script)
+//        markdown = Down.init(markdownString: script)
+//        if let converted = try? markdown.toAttributedString(.default, stylesheet: Issue.style) {
+//            attributedScript = converted
+//        }
+
```
subdigital: +1 on deferring markdown rendering until you've reached that screen. We can delete these comments.

FiveCalls/FiveCalls/IssueDetailViewController.swift
```diff
@@ -174,7 +175,14 @@ extension IssueDetailViewController : UITableViewDataSource {
             
         case IssueHeaderRows.description.rawValue:
             let cell = tableView.dequeueReusableCell(withIdentifier: R.reuseIdentifier.descriptionCell, for: indexPath)!
-            cell.issueLabel.text = issue.reason
+
+            let markdown = Down.init(markdownString: issue.reason)
```
subdigital: `Down.init` should just be `Down` ya?

FiveCalls/FiveCalls/Info.plist
mccarron: Yup the submission warned me so I should have pushed that change afterwards.

FiveCalls/FiveCalls/IssueDetailViewController.swift
mccarron: I just like the `Down in it` usage. ;) 

But yeah bad habits, will fix in my second pass tonight. I'll try to hook up the UITextView so we can handle links as well.

nickoneill:
I swapped out the label for a textview here, but now the textview has to grow to be as large as the content which was stretching the limit for time I could apply to this at the moment.

Sorry for the only partially useful commit - let me know if I'm not being helpful 🤪

mccarron:
Looks good to me, but I can't approve as its my PR.

nickoneill:
Naturally, as soon I want to check what the test failure was, Circle CI is having a site-wide outage.

nickoneill:
Ah, this is the screenshot UI tests failing. Does Circle support this or should we specifically exempt it from the tests that are run? @mccarron you're the circle expert, yeah?

mccarron:
I can dig in on Sunday to disable that. It's useful locally only really.

nickoneill:
That would be awesome @mccarron, as far as I can see this is the last item on our list for release!

mccarron:
Tests now run and pass! 🤘

PR #232, opened 2018-09-05
Token refresh on stats 401

subdigital:


subdigital:
@nickoneill not really sure what would cause these to get stuck -- I've observed credentials get renewed and saved correctly. We are logging and ignoring errors here, so I've added a thing that will send this to Crashlytics -- not sure if this is overkill, but could give us an idea how often these failures occur.

subdigital:
@nickoneill ok, added a refresh mechanism if we see a 401 from the server. Tries to refresh twice, if we still get a 401 it will fail.

PR #236, opened 2018-09-06
Use the server's value for streaks if available

nickoneill:
Two new values from the stats endpoint:

* The number of streak weeks
* The time of the first recorded call for this user

We don't currently use the second in iOS, but on the web it powers a field that says "Making calls since May 2017" or some such.

### DIFF
--- a/FiveCalls/FiveCalls/MyImpactViewController.swift
+++ b/FiveCalls/FiveCalls/MyImpactViewController.swift
@@ -15,6 +15,7 @@ class MyImpactViewController : UITableViewController {
     
     private var viewModel: ImpactViewModel!
     private var userStats: UserStats?
+    private var weeklyStreak: Int?
     private var totalCalls: Int?
     
     @IBOutlet weak var navSignInButton: UIBarButtonItem!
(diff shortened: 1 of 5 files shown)


FiveCalls/FiveCalls/MyImpactViewController.swift
```diff
@@ -103,14 +104,19 @@ class MyImpactViewController : UITableViewController {
     
     func displayStats() {
         viewModel = ImpactViewModel(logs: ContactLogs.load(), stats: userStats)
-        let weeklyStreakCount = viewModel.weeklyStreakCount
+
+        // use the server weekly streak if we have it
+        let weeklyStreakCount = self.weeklyStreak ?? viewModel.weeklyStreakCount
         streakLabel.text = weeklyStreakMessage(for: weeklyStreakCount)
```
subdigital: all of this should be in `ImpactViewModel`, including this `weeklyStreakMessage`. The result should just be 

```swift
streakLabel.text = viewModel.streakMessage
```

nickoneill: 👍

nickoneill:
OK, I moved the weekly streak count into the UserStats object as well so the view model can continue being initialized with the same parameters.

PR #243, opened 2018-10-08
Align Gemfile.lock & Podspec.lock cocoapods version to 1.5.3

iainsmith:
Just checking out the project and noticed that the bundled version of Cocoapods is 1.5.0
where as the [Podspec.lock](https://github.com/5calls/ios/blob/92a84c7dd1e6004a4ee46c9312e46fe5d78af7e7/FiveCalls/Podfile.lock#L90) version is 1.5.3 on master.

Seems sensible to bump the bundled version to 1.5.3, unless anyone has any objections.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 source "https://rubygems.org"
 
-gem "cocoapods"
+gem "cocoapods", "1.5.3"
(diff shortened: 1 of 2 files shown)


Abizern:
Looks good to me. Aligns to what is actually used. Thanks @iainsmith 👍

PR #244, opened 2018-10-09
Supress warnings for Down & Auth0

iainsmith:
This PR tells cocoapods to supress warnings from Down & Auth0, as they were adding
unnecessary warnings in Xcode.

This is a bit of a noisy diff, but it's all generated from the Podfile change & running `bundle exec pod install`.


nickoneill:
I didn't know you could do this in the podfile, thanks @iainsmith!

Abizern:
I had no idea either. Thanks @iainsmith !

Issue #144, opened 2017-04-02
Support Dynamic Type Changes


subdigital:
Currently the app doesn't respond to dynamic type changes. This is something we should fix to be more accessible.

A probably incomplete list of places that need this work:

- [x] Issue Detail
- [x] Call Scripts
- [x] About screen
- [x] Onboarding screens
- [x] Issue list (front page and more)
- [x] Impact
- [x] Reminders
- [x] Set location

Abizern:
Related to this - How tied are we to using Roboto rather than the system font? This would be much smaller task if that were the case.

I realise that there may be issues about maintaining a consistent typeface across Web, Android and Mobile. I also realise that it is something of a personal bugbear of mine to use the Android system font on an iOS device.

Of course, this would mean making other related changes to the app and the css files used for serving stored web content. But the app is already a little inconsistent with the Welcome screen using the system font.

So it would make implementing this feature easier, and make the app look more native, at the cost of having to rework parts of the app and, maybe the design.

nickoneill:
I'm happy to use the default for the platform for most text, particularly if it makes things simpler. The only case I really care about maintaining the font is in the logo which is generally not dynamically rendered anyway. We didn't pick Roboto (and technically it's Roboto Condensed) based on android, it was inherited from the logo usage when we were web only.

Abizern:
I don't want to mess with the logo, and that's not something that's affected by this.

If a few more people come in on the same side I'll start work on it in a couple of days (unless someone else wants to take it). It's a good step for accessibility.

subdigital:
+1 on native font

iainsmith:
Hey All,

I've started diving into this. As we're supporting iOS 10 we can mostly get away with using 

`adjustsFontForContentSizeCategory = true // or set in storyboard`
`font = UIFont.preferredFont(forTextStyle: .body) // or set in storyboard`

I have a branch on my fork that I'll open up a WIP PR for.

# Screens updated

I'm going to leave a list of areas here to remind myself.

### IssueDetailViewController

* [x] CSS support for dynamic type
* [ ] Representatives title
* [ ] Representatives details

nickoneill:
Do your two PRs handle all the cases for dynamic type @iainsmith?

iainsmith:
Hey @nickoneil, there are a few screens that don’t have any support for dynamic type yet. We could leave this ticket open for a while or open new tickets that are specific to a screen.

nickoneill:
[...]

PR #245, opened 2018-10-09
Dynamic type support - Issue Detail & Call Script

iainsmith:
**Updated 10 Oct 18**

This PR adds basic Dynamic Type support (#144) to the IssueDetailViewController & CallScriptViewController.

## Approach

The changes for full Dynamic Type support are going to land up touching each screen so I'd suggest we break this into multiple PR's if that sounds good with everyone? There are a lot of Gotchas when trying to support the Extra large AX1 - AX5 sizes so I'd suggest we break this down as follows:

* Pull Request into master for 1-2 screens at a time
* First iteration -> Best effort to support xSmall -> xxxLarge with minimal refactoring. (Largely storyboard changes)
* Future iterations -> Improvements to AX1-AX5 font sizes or that require larger code changes (Moving to StackViews, etc)
* I'd like to suggest we ignore visual defects that:
    * are only reproducible when changing Dynamic Type while a ViewController is on screen. This is such a low usage scenario that it’s not worth focusing on imho. 
    * only occur at the extra large sizes AX1-AX5 for the first few passes.

Does that sound like a reasonable approach to everyone?

## Changes in the PR

|Screen | small | xxxLarge |
| --- | --- | --- |
| Issue Detail | detail-small | detail-xxxlarge |
| Issue Detail + profile | detail-profile-small | detail-profile-xxxlarge |
| Call Page | call-small | call-large |

## Deferred to later

* The Call Script page uses a collectionView for the dynamic buttons. To support dynamic height here we'd either have to look at resizable CollectionViewCells or potentially move to a simpler Stack View layout that can easily switch between a grid and a vertical stack.
* The Call Script header breaks down at AX2

### DIFF
--- a/FiveCalls/FiveCalls/Base.lproj/Main.storyboard
+++ b/FiveCalls/FiveCalls/Base.lproj/Main.storyboard
@@ -381,7 +381,7 @@
                                             <rect key="frame" x="0.0" y="0.0" width="320" height="122"/>
                                             <autoresizingMask key="autoresizingMask"/>
                                             <subviews>
-                                                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="SPEAK OUT FOR ENVIRONMENTAL IMPACT AT STANDING ROCK" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="IJo-u6-qUp">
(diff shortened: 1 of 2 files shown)


Abizern:
As it's a WIP branch it's maybe not mergeable yet, but looks good so far, I think

mccarron:
Looks like a good starting point, keep it up. Definitely try to expand labels where you can, but curious what we should do about the buttons. 🤔

mccarron:
Thanks for updating the screens, I think it looks great, lets merge it in.

nickoneill:
Thanks @iainsmith! This looks great, I really appreciate the visual examples

Issue #253, opened 2018-11-08
Issues are in the incorrect order on iOS


nickoneill:
We just posted a new call and it appears near the bottom of the results on iOS. Previously (at least I'm pretty sure?) the order was the same as the website.

As of yet, the order is determined only by the order in which it appears in the json, we have a proper ordering value on the server side though that we could pass through to clients if it makes more sense.

<img width="383" alt="screen shot 2018-11-08 at 9 14 30 am" src="https://user-images.githubusercontent.com/49688/48215280-c5bc6f00-e336-11e8-9635-e36afb3884fd.png">

vs

![img_3541](https://user-images.githubusercontent.com/49688/48215326-dec52000-e336-11e8-8584-20d5b9bc186e.jpg)

mccarron:
I can look tomorrow but can't today. Have we confirmed the JSON order is the same?

nickoneill:
Yep, and the order on the website follows the JSON. For example right now the top 5:

```
      "id": "recBJ6wBlShm6b5z4",
      "name": "Demand Acting AG Matthew Whitaker Recuse Himself from Mueller Investigation",
      ...
      "id": "recFS1sCVCQhNqZUt",
      "name": "Protect Mueller and the Russia Investigation",
      ...
      "id": "recgvn5j8YZFeEytQ",
      "name": "Demand Action Against Gun Violence",
      ...
      "id": "recgUbNBp4Idrz6nk",
      "name": "Pass the Violence Against Women Reauthorization Act of 2018",
      ...
      "id": "recSQrrm6a0RR6dbK",
      "name": "Demand the Treasury Department Release Trump's Tax Returns",
      ...
```

These are mixed in the app, some at the bottom, some closer to the top of the list.

mccarron:
Ah it's a Swift 4.2 change to how Hashes are stored, it's now randomly seeded so you can't expect the input order to be maintained. 

So we will need to sort the returned elements because now you can't count on the order of Dictionary objects.

nickoneill:
Can we avoid storing the issues list as a dictionary? Arrays still have consistent ordering, right?

Abizern:
Unfortunately I don't have Xcode 9 anymore so I can't load the current code. Looking at the result of the call to get issues, though it seems  we use the "issues" key which returns an array already.

subdigital:
[...]

PR #255, opened 2018-11-21
Sort Issues by Order Field

mccarron:
This will sort issues by the JSON order to avoid the list being out of order on the new Swift compiler. 

This addresses issue #253 

simulator screen shot - iphone xs max - 2018-11-20 at 18 28 45


nickoneill:
Looks great, thanks for putting this together so fast!

Issue #256, opened 2018-12-03
Experiment with mixpanel for user behavior analytics


nickoneill:
As part of #239, we're replacing fabric with something else for mobile analytics. Before we go full-on tracking everything in the app, I'd like to see how it behaves tracking data on multiple platforms.

I've already integrated mixpanel with the website. I'm going to do a quick test integration for iOS on 1.5.3 so we can see if both sets of data can live nicely along side one another.

nickoneill:
we experimented, it's garbage

Abizern:
There are still a couple of calls that use it, and the Pod. Can they be removed?

nickoneill:
Yep, I'm still looking for a good way to get insights into how many users we have and how they use the app, but we can remove mixpanel because I'm not using it at all right now.

PR #257, opened 2018-12-03
Added experimental mixpanel integration

nickoneill:
For #256, last thing to do before 1.5.3 goes out.


FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -41,6 +42,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
 
         oneSignalStartup(launchOptions: launchOptions)
+        Mixpanel.sharedInstance(withToken: "776fce75f7e3ddfbb13b615dcb94ff95")
```
subdigital: Do we already have a place to store this stuff? I don't like committing tokens in the source when I can avoid it.

FiveCalls/FiveCalls/IssueDetailViewController.swift
```diff
@@ -27,6 +28,8 @@ class IssueDetailViewController : UIViewController, IssueShareable {
         tableView.estimatedRowHeight = 100
         tableView.rowHeight = UITableViewAutomaticDimension
         NotificationCenter.default.addObserver(self, selector: #selector(madeCall), name: .callMade, object: nil)
+
+        Mixpanel.sharedInstance()?.track("Topic", properties: ["IssueID": self.issue.id])
```
subdigital: What would you say to an extension on `UIViewController` like this:

```swift
extension UIViewController {
    func trackEvent(_ name: String, properties: [String:String]) {
        Mixpanel.sharedInstance()?.track(...)
    }
}
```

subdigital: This way we decouple the actual type (and service) used to track these events and cleans up the callsite a bit.

FiveCalls/FiveCalls/IssuesViewController.swift
```diff
@@ -56,7 +57,12 @@ class IssuesViewController : UITableViewController {
         self.registerForPreviewing(with: self, sourceView: tableView)
 
         Answers.logCustomEvent(withName:"Screen: Issues List")
-        
+        if self.shouldShowAllIssues {
+            Mixpanel.sharedInstance()?.track("More")
+        } else {
+            Mixpanel.sharedInstance()?.track("Home")
+        }
```
subdigital: With the above change, we could extract this into a single call:

```swift
var analyticsEvent: String {
    if shouldShowAllIssues { 
        return "More"
    } else {
        return "Home"
    }
}

...

trackEvent(analyticsEvent)

```

FiveCalls/FiveCalls/IssueDetailViewController.swift
nickoneill: 👍

FiveCalls/FiveCalls/AppDelegate.swift
nickoneill: Info.plist seems like what we've been using. I'll pop it in there.

nickoneill:
Updated!

FiveCalls/FiveCalls/IssueDetailViewController.swift
```diff
@@ -10,6 +10,7 @@ import UIKit
 import CoreLocation
 import Crashlytics
 import Down
+import Mixpanel
```
subdigital: don’t need this in your view controllers anymore, since you’re calling an extension method.

subdigital:
Otherwise looks good

nickoneill:
Updated and rebased, will merge after the checks complete.

PR #258, opened 2018-12-03
Log any error we get back from stats

nickoneill:
...in an attempt to get more info on one users's weird error


FiveCalls/FiveCalls/MyImpactViewController.swift
```diff
@@ -156,6 +156,9 @@ class MyImpactViewController : UITableViewController {
             // Fetch the user's call stats
             let userStatsOp = FetchUserStatsOperation()
             userStatsOp.completionBlock = {
+                if let error = userStatsOp.error {
+                    Answers.logCustomEvent(withName: "userstats-error", customAttributes: ["message": error.localizedDescription])
```
subdigital: Can we use Mixpanel hear too?

nickoneill: We can - though I feel like we're misusing either tool using it like this. Is there a generic error logging tool that would do a better job for it?

nickoneill:
Looks amazing - I just applied. Looks like they support JS error tracking too, so we can drop sentry and unify errors in one place!

nickoneill:
Bugsnag integrated! Looks like it's much more well suited for this. I'm going to limit our integration to this one area until we a) hear back about the open source plan and b) get some real data into the dashboard so we can evaluate how useful it is.

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -62,6 +64,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
     }
 
+    func bugsnapStartup() {
```
subdigital: bugsnap?

subdigital:
I have similar comments about not letting bugsnag leak everywhere in every file, and use an extension method to keep that implementation detail in one place, however I hear your point about trying it out. If we end up committing to it we can move this logic.

FiveCalls/FiveCalls/AppDelegate.swift
nickoneill: lol, fixed

PR #261, opened 2019-01-30
Use of New API

subdigital:
Paired with @mccarron  on this one.

This replaces our previous API with the new one.

Notable changes:

- Removed custom JSON parsing, now we use `Codable` models
- Issues no longer contains an array of Categories, instead we leverage [CategorizedViewModels]
- Because of the above, we need to fetch contacts for the main screen (for stats) _and_ the issue detail screen. This could definitely be improved (maybe once at launch, refetched when user location changes)
- Got rid of a flicker in IssueDetail coming back from Call Script VC, but this reveals a latency issue w/ the Markdown parsing... I never noticed this before but it can take upwards of a second to load. This should probably be tested on device to see if it's a real issue and if we want to add some sort of loading state for the issue text.

Still to figure out:

- Split districts


nickoneill:
The amount of code deletion here is excellent!

* IIRC we ran into the markdown loading time last release and decided it was negligible on the device.
* Can you explain a bit more why contacts should be loaded twice? Since we deliver a payload of every contact type we support, you can request it once and then filter that local list for any subsequent issues by the `contactArea`.

I'm booting (ok, charging...) my android device right now to give an example of how katie handled the new split district stuff. We'll be doing something similar on the web very soon.

nickoneill:
It looks like this is mostly complete @mccarron:

- [ ] Would like to better understand why contacts need to be fetched twice (though if it's easier for you, it doesn't put significant additional strain on the server)
- [ ] Select the first house rep in the `house` array for display (it's the most likely one, and I'll be trialing a UI fix for selecting which one is correct on the web first)
- [ ] And if you can, check out if #262 is an easy fix

nickoneill:
OK, I see now why you needed the contacts for the issue list page too!

I updated this for Swift 4.2 (because I was initially working on something Hashable related) and added a small contact cache so we can fetch contacts as much as we want without waiting each time. Also updated to select only the first House contact.

Can I get a review on this @mccarron or @subdigital?

FiveCalls/FiveCalls/CategorizedIssuesViewModel.swift
```diff
@@ -0,0 +1,33 @@
...
+        return category.name
+    }
+
+    init(category: Category, issues: [Issue]) {
+        self.category = category
+        self.issues = issues
+    }
+}
+
+extension CategorizedIssuesViewModel : Equatable, Hashable {
```
Abizern: I don't think you need to specify `Equatable` here as `Hashable` descends from it.

FiveCalls/FiveCalls/CategorizedIssuesViewModel.swift
```diff
@@ -0,0 +1,33 @@
...
+    static func == (lhs: CategorizedIssuesViewModel, rhs: CategorizedIssuesViewModel) -> Bool {
+        return lhs.category == rhs.category
+    }
+
+    var hashValue: Int {
```
Abizern: This method on `Hashable` is deprecated. You might want to consider replacing it with:

    func hash(into hasher: inout Hasher) {
        hasher.combine(category)
    }

FiveCalls/FiveCalls/Category.swift
```diff
@@ -13,55 +13,12 @@
...
-}
-
-extension Category: Hashable {
-    var hashValue: Int {
-        return name.hashValue
-    }
 }
 
-extension Category: Comparable {
+extension Category : Equatable, Hashable, Comparable {
```
Abizern: Same as previous comment about `Equatable` not being needed.

FiveCalls/FiveCalls/Extensions/Error+Offline.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+
+extension Error {
+    func isOfflineError() -> Bool {
+        let e = self as NSError
+        guard e.domain == NSURLErrorDomain else { return false }
+        
+        return e.code == NSURLErrorNetworkConnectionLost ||
```
Abizern: Just a style comment (you may or may not agree), but I find that 

    return [NSURLErrorNetworkConnectionLost, NSURLErrorNotConnectedToInternet, NSURLErrorSecureConnectionFailed].contains(e.code)

Is less code and more readable.

nickoneill:
Thanks @Abizern! I'll clean up the hashable bits later today

FiveCalls/FiveCalls/Extensions/Error+Offline.swift
nickoneill: Agreed, I find this style much more readable

nickoneill:
Just retested with split districts and changing locations, happy with the current state of things and getting the 4.2 update out of the way will unblock some other work.

Thanks @mccarron @subdigital and @Abizern for your work on this!

PR #267, opened 2019-06-16
Convert to UNUserNotifications

### LINKED ISSUE #240: Migrate to UserNotification framework
Using a bit of LocalNotification and the old UIUserNotificationSettings stuff for local and remote notifications. It's deprecated since iOS 10, we should start moving away from it.

nickoneill:
Existing user notifications are carried over! That makes things easier. Fixes #240 

This does the very basic conversion to the non-deprecated API. We should probably allow reminding on weekends in the future too.

### DIFF
--- a/FiveCalls/FiveCalls/ScheduleRemindersController.swift
+++ b/FiveCalls/FiveCalls/ScheduleRemindersController.swift
@@ -7,6 +7,7 @@
 //
 
 import Foundation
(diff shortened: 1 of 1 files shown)


nickoneill:
Sorry, gonna merge this soon but I want to do a day of testing on my device and had a busy weekend :)

PR #274, opened 2019-12-22
Removed pantry pod and codabled everything

### LINKED ISSUE #182: Remove Pantry pod
Depends on #179.

The project uses Pantry for deserialisations and storage. As noted in Pantry's README, it is no longer going to be supported.

So we should probably refactor to use the Codable protocol available in Swift 4. Not a small job.

nickoneill:
Only thing that needs to stick around is a new wrapper struct that we discard. Fixes #182


nickoneill:
Not entirely sure why the tests are failing on circle, they seem to build OK locally.

nickoneill:
Testing via fastlane fails too. It appears that `NSSearchPathForDirectoriesInDomains(.applicationSupportDirectory, .userDomainMask, true)` isn't a good choice for headless testing.

Not a huge fan of how I use a different path when detecting tests but it seems to work OK. I tried creating a directory at the application support path if it didn't exist but that also failed.

FiveCalls/FiveCalls/ContactLog.swift
```diff
@@ -93,10 +93,16 @@ struct ContactLogs {
...
         let pantryDirName = "com.thatthinginswift.pantry"
         let appSupportDir = NSSearchPathForDirectoriesInDomains(.applicationSupportDirectory, .userDomainMask, true).first!
-        let targetPath = URL(fileURLWithPath: appSupportDir).appendingPathComponent(pantryDirName).appendingPathComponent(ContactLogs.persistenceKey, isDirectory: false)
+
+        let targetPath: URL
+        if UserDefaults.standard.bool(forKey: "isBeingTested") {
+            targetPath = FileManager.default.temporaryDirectory.appendingPathComponent(ContactLogs.persistenceKey, isDirectory: false)
+        } else {
+            targetPath = URL(fileURLWithPath: appSupportDir).appendingPathComponent(pantryDirName).appendingPathComponent(ContactLogs.persistenceKey, isDirectory: false)
+        }
```
subdigital: Funny enough I did a screencast on how I tend to do this sort of thing:

image

https://nsscreencast.com/episodes/341-detect-testing-in-app-delegate

nickoneill: Much better - less configuration is better.

Issue #233, opened 2018-09-05
Call results are double-submitted


nickoneill:
It looks like we're double-counting calls when people are logged in on iOS. See this view into the database:

<img width="648" alt="screen shot 2018-09-05 at 1 43 17 pm" src="https://user-images.githubusercontent.com/49688/45120108-108cf100-b112-11e8-970f-a7ced9366127.png">

Since we're still keeping local records of call results, it looks like we're creating them without marking them as already uploaded. The backfill task then tries to upload them again.

Sending the results to `/report` will connect the result to the user already, we don't need to upload them again via `/users/stats`

nickoneill:
(I also see a record of this *not* happening with logged in users on iOS, so it might be more complicated than just "we don't mark them as uploaded")

PR #276, opened 2019-12-22
Wait a few minutes before backfilling unreported stats

nickoneill:
It's possible that #233 was caused by network races - as the report is going back, we'll see if unreported stats need to be filled in the `viewDidLoad` for the main issues screen. Instead of reporting them immediately, let's only report backfilled stats that are more than a few minutes old.

Additionally, we needed to update auth0 for iOS 13, otherwise it would break login.


subdigital:
👍🏼

PR #290, opened 2020-01-13
Remove promisekit in favor of some local Results

nickoneill:
I find it really hard to reason about promises when implementing them in code, using them is clearly superior but since we're actually implementing them here it's still confusing. This work is simple enough that using `Result` works fine.

This is based on the swift 5 branch, need to fix the tests and merge that first.


nickoneill:
I removed some of the mystery for me when refactoring this... I think if I reduce the promise jargon it's more understandable. I'll take another pass at it by modifying it a bit rather than removing it.

Issue #282, opened 2019-12-25
Add donations to the app


nickoneill:
Prompt users to donate to 5 Calls at some appropriate time. We already prompt for ratings and notifications after an early set of calls are done, so this can possibly be an ask after a larger volume of calls. Put it in the about page as well.

Apple has a non-profit approval process for something - maybe less fees - but last I checked we don't qualify since we're not a c3. I hate to give Apple 30% of whatever people want to donate to us but if we make it subscription-based then at least after a year we will be at 15%.

Least effort and lowest odds of approval: we open a web view with our actblue page.

More effort: use something like revenuecat to make monthly and yearly subscriptions.

nickoneill:
This is going to be a cell that shows up in the list after you've made some calls. It'll have two buttons, "donate now" and "I've donated" and both will make the cell disappear and take you to our actblue page in safari.

Cell should show again in ~3 months.

Issue #291, opened 2020-01-25
New "done" page with results, tally, share


nickoneill:
After a user has clicked results for all their contacts, push a new screen that has their results nicely summarized, the total calls and a share photo they can use.

![Contacts redesign](https://user-images.githubusercontent.com/49688/73127085-0e7a3b00-3f70-11ea-91cd-520dbbe0f329.jpg)

PR #292, opened 2020-02-01
Rework call script ends for done page

nickoneill:
Rather than end the script flow with a checkmark and have to go back through all the contacts, we have a new done page that shows the results of your calls, call totals, a donation link (for people who have made more than a few calls) and a share preview.

Prev:
Screen Shot 2020-01-21 at 1 52 19 PM

Now:
Screen Shot 2020-02-01 at 12 47 44 PM

* this also includes removing the kingfisher pod for a basic image extension to fetch images from urls. I like to replace our limited usage of 3rd party deps with one file of code.

To fix before merge:
- [x] Add rating and notification prompt to done page
- [x] Check iPad support


nickoneill:
Fixes #282 and #291

PR #296, opened 2020-02-12
Fix crash reporting

subdigital:


nickoneill:
Huh, maybe this was one of those things I set up in the branch that I discarded. Thanks for adding!

PR #298, opened 2020-03-07
If a contactmanager doesn't exist, create one

nickoneill:
Trying to fix this crash, where contactmanager is nil:

https://appcenter.ms/orgs/5-Calls/apps/5-Calls-iOS/crashes/errors/769878387u/overview


nickoneill:
This is just to ensure that the `ContactsManager` isn't nil before we use it. There's no (significant) time spent on init, so nothing to promise on until `fetchContacts` happens below the addition.

Issue #278, opened 2019-12-24
Support dark mode


nickoneill:
I've forced the style to light for 1.6 release, but we should really support dark mode.

PR #311, opened 2020-07-07
Dark mode for Scheduling reminders

Abizern:
Dark mode for enabling reminders.

Partly addresses #278

I think that's it for the initial pass over the About.Storyboard.

Haven't made changes to the Open Source acknowledgements as the controllers for that come from a Pod and I don't want to go into that.

Sorry about the small PRs. I keep thinking I'm done for the day and then do a little bit more.

### DIFF
--- a/FiveCalls/FiveCalls/Base.lproj/About.storyboard
+++ b/FiveCalls/FiveCalls/Base.lproj/About.storyboard
@@ -219,16 +219,16 @@
                                 <userDefinedRuntimeAttributes>
                                     <userDefinedRuntimeAttribute type="string" keyPath="titlesString" value="Mon, Tue, Wed, Thur, Fri"/>
                                     <userDefinedRuntimeAttribute type="color" keyPath="selectedBackgroundColor">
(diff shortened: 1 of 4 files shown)


Abizern:
I'm making lots of small changes, Moving colors to the Asset catalog and using them in the storyboards. Once it's all done then a more detailed UI pass can be made and the colors in the catalog can be tweaked to look their best on light or dark mode.

Abizern:
Reminder-None-Light
Reminder-Chosen-Light
Reminder-Empty-Light

Abizern:
Reminder-Chosen-Dark
Reminder-Empty-Dark
Reminder-None-Dark

nickoneill:
Sorry for the wait on this, will review this weekend

nickoneill:
Appreciate the work moving these to the color catalog. I'm going to take a stab at mocking something that has better readability than blue-on-dark-grey but we can continue making these improvements on the color handling in parallel.

Abizern:
I've been a bit lazy with naming the colours. Semantic labelling would be better and maybe even provide cognitive freedom to use different colours entirely.

PR #319, opened 2021-09-30
Removed auth0 in favor of firebase

nickoneill:
Auth0 auth has always been a thorn in my side. Now the web runs anon firebase auth and I extended it here so we can get a unified view of all users. I don't like google either but this sure is less work.

Removed: Auth0, AppCenter
Added: Firebase auth, non-IDFA analytics, crashlytics

If it's important to people in the future, maybe we can add actual signin methods so people can sync their progress across web and mobile, but for now I don't think it's necessary.


nickoneill:
Fixed the bitrise build with some magic, disabled circleci since it's not needed anymore

PR #320, opened 2022-06-04
Ditch pod in favor of spm, resolve warnings, update some deps, etc

nickoneill:
Diff is crazy, but very little code changes other than upgrading onesignal to v3.


Abizern:
There seems to be something weird with this. I can't even build this branch locally (Even after adding a dummy GoogleService-Info.plist). Building and running fastlane tests on master does work, so there may be an issue with what you have locally in your branch that isn't matching what's in the repo and causing build and test failures.

nickoneill:
There's a note in the Readme to vendor the rswift binary as `vendor/rswift`... perhaps we should just include it though! Or at least loudly complain if it's not there.

nickoneill:
The `xcodebuild showBuildSettings` takes a bit with SPM, looks like we set an env var to extend the timeout on bitrise previously but the env var changed at some point... updated it and running again.

nickoneill:
Fixed on bitrise! LMK if it works for you now @Abizern

Abizern:
> Fixed on bitrise! LMK if it works for you now @Abizern

The fastlane script works, and the building locally works (once I pull in the dummy info.plist) 👍 👏

Abizern (approved):

PR #321, opened 2022-07-24
Fixed a bug where contacts were not loaded on the more page

### LINKED ISSUE #316: All issues list doesn't display progress on completion circles
A user reports that issues on the "all issues" list doesn't report completion % or all completed in the list. It works fine on the main list. I can reproduce it on my device.

nickoneill:
User reported that the progress indicators were not filled after making calls. Turns out this was two issues:

* after Done page when user navigates back to the main list, progress was not showing
* contacts were not loaded on the More list because we already have a list of issues, so progress was not showing

fixes #316


Abizern (approved):

nickoneill:
Thanks for looking @Abizern!

PR #323, opened 2023-06-18
Prep for some attribution work

nickoneill:
Create an anonid which can tie together push notifications and call reports. Set the anonid and a generalized district in OneSignal for targeting, and slightly reword the push prompt to better explain our goals.


Abizern (approved):

nickoneill:
thanks @Abizern !

PR #324, opened 2023-06-29
Save the firebase id for when we remove firebase later

nickoneill:
Gonna remove firebase at some point but we'll do a release that saves the userID before then so when we move away we can still use it.

### DIFF
--- a/FiveCalls/FiveCalls/SessionManager.swift
+++ b/FiveCalls/FiveCalls/SessionManager.swift
@@ -20,6 +20,12 @@ class SessionManager {
     func startSession() {
         Auth.auth().signInAnonymously { result, error in
             self.userID = result?.user.uid
+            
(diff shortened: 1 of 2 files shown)


Abizern (approved):

Abizern (approved):

Abizern:
If you're planning to remove Firebase, what are you going to use to use for authentication on requests?

nickoneill:
While we do send the auth token and validate it on the server, adding firebase was more about unified analytics than auth. If we trust that a UUID that a user sends is correct, worse case scenario someone brute forces a UUID (unlikely) and gets access to a summary of call result counts (unavailable, voicemail, etc) you made, but not potentially sensitive information like a location or even a congressional district.

I don't think we really need it for what we want to do, and if we decide we need it in the future I would much prefer doing a lighter auth layer than the huge dependency graph firebase brings with it.

Abizern:
Thanks. Good to know.

PR #325, opened 2023-07-04
Add some pages to plausible and move some old analytics out of the way

nickoneill:
Another step closer to no firebase. My first attempt at a swift package too.


Abizern (approved):
I've got some clean-up suggestions, but I'll raise that in an subsequent PR on top of this one.

PR #328, opened 2023-07-07
Combine properties so the compiler is happy

nickoneill:
`DEBUG` flag hid a compiler error for this one :)


Abizern (approved):
Good catch!

PR #329, opened 2023-07-22
Start a SwiftUI interface

### LINKED ISSUE #306: iOS 14: One-time Location
We're a perfect fit for one of the latest improvements in privacy for iOS 14: one time use location prompts. We only get your location once when setting it and then we never use it again unless the user needs to change their location so providing access whenever the app is open is excessive.

* If a user is on iOS 14, when getting location automatically prompt for one-time access only.
  * Users on iOS 13 and before should get the old behavior.

### LINKED ISSUE #331: SwiftUI design / 2.0 release
App needs a design refresh, both because it's been 6 years since the initial work was done and because SwiftUI makes it a hell of a lot easier to build new stuff.

Started on quite a bit of it here: https://github.com/5calls/ios/pull/329
I would like to not keep the whole redesign in a single PR but unless there are other changes hitting main it doesn't really matter. 

**Tracking some (not at all comprehensive) todos:**
- [x] Loading issues, contacts
- [x] Send outcomes back to server
- [x] Setting location, persisting location
- [x] Dashboard / Main issues view
- [x] #352
- [x] Issue detail view
- [x] Issue contact detail view (i.e. script and phone numbers for an issue + specific contact)
- [x] Done page design
- [x] About view
- [x] Reminders view
- [x] My Impact view
- [x] #349
- [x] #395
- [x] #365
- [x] Modern welcome screen @heathermh
- [x] #363
- [x] #366
- [x] #377
- [x] #379
- [x] #400
- [x] #380
- [x] #388
- [x] #387
- [x] #384
- [x] #409
- [x] for contact's list of phone numbers, can we keep the phone symbol showing at larger dynamic type settings? https://github.com/5calls/ios/pull/427
- [x] check if IssueContactDetail needs to check for reduce motion / limit animations (I don't think it does)

**Bugs**
- [x] Welcome screen doesn't show paging pips
- [x] #362
- [x] Placeholder contact circles are not aligned correctly
- [x] #368
- [x] If we haven't loaded a contact for an area, clicking "see your script" crashes (we should probably show an error "can't find any contacts for this issue in your location, try a different location")
- [x] ~~iPad: tapping twice on "Detect my location" complains about leaked continuations (my iPad takes a long time to get a location? needs a disabled state)~~ seems fine now
- [x] iPad: changing your location should reload the issue detail to reflect new contacts
- [x] #382
- [x] Placeholders show 2 for House, 1 for Senate, should be reversed
- [x] ~~Location header on iPad cannot be tapped~~ dunno, this seems fine now
- [x] #391
- [x] #401
- [x] #393
- [x] #413
- [x] #426
- [x] #428 
- [x] https://github.com/5calls/ios/issues/430

**Cleanup tasks for after this is done and ready for prod:**
- [x] Move plausible tracking to the new views
- [x] #370
- [x] Remove unneeded 3p stuff: firebase, down, Promises

nickoneill:
Started working on a refresh to the interface in SwiftUI, no data connected yet. I would like to not have a huge branch with a complete rewrite so this switches on a key in AppDelegate to show the existing app or the new swiftUI stuff.

Also bumps the min iOS version to 16... I would like to reduce that to 15 but I'll revisit when it's closer to done to see what fun features we would have to stop using.

Fixes #306 
Fixes #331


Abizern:
This is going to take some thought...

nickoneill:
Added a first approach for loading and storing app state, kindof a naive approach for what little I know about data flow in swiftui

nickoneill:
Inspired to refactor what the issuesmanager knows about, moving it from updating the app delegate itself to calling a function on an arbitrary delegate. Not even sure we need an issues manager if we do things this way... maybe just one operation executor that knows how to run all the various operations?

nickoneill:
Over the weekend I did some work to improve the display of reps with no images, removed the issue list item checkmark (for now... we still need a way to represent that an issue has been called) and made some headers for setting your location

nickoneill:
Last night I added persistence to user locations, initially I was working with the existing `UserLocation` model but the way we structured it previously with a singleton was not very swiftui-y so I copied the relevant model bits to `NewUserLocation`. A little more complicated than using `@AppStorage` because we have a legacy format to deal with, but I think it is integrating nicely with our `AppState` model

nickoneill:
Added some work on the design for an issue detail page. Not hooked up to navigation yet.

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -0,0 +1,91 @@
...
+                        if case let .serverError(error) = result {
+                            print("issues error: \(error)")
+                        }
+                        if case .offline = result {
+                            print("issues offline")
+                        }
+                    })
+                }
+        
+                if appState.contacts.isEmpty && appState.location != nil {
```
cselin, line 57:
```suggestion
                if let location = appState.location && appState.contacts.isEmpty {
```

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -0,0 +1,91 @@
...
+                    op.fetchContacts(location: appState.location!, delegate: (UIApplication.shared.delegate as! AppDelegate)) { result in
```
cselin, line 58: If using suggestion above:
```suggestion
                    op.fetchContacts(location: appState.location, delegate: (UIApplication.shared.delegate as! AppDelegate)) { result in
```

cselin (commented):
Great progress so far! I haven't fully viewed everything yet, but wanted to leave some comments in the meantime.

A few general suggestions that aren't necessary, but would help moving forward:

1. Support only iOS 16+ for the SwiftUI version from the start instead of including iOS 15. That way we can avoid AdaptiveSheet entirely, start with `NavigationStack` instead of `NavigationView`, etc.
2. Consider adapting unidirectional data flow with a central store instead of continuing to pass around the app delegate. I don't think this would be a ton of work to implement, but also could be something we do later since you already have things functional with the delegate. We could also potentially remove the necessity for the app delegate completely by using ScenePhase.
3. Replacing completions with async/await and Combine would be nice-to-haves, but by no means as helpful as the above.

cselin:
Also found a few bugs (I'm sure there a plenty since this is an in-progress migration, but wanted to capture them here:

1. I don't see any page indicators on the intro view Simulator Screenshot - iPhone 14 Pro - 2023-09-06 at 15 38 07
2. Consider using safe area for cleaner scrolling and it looks like there's some funky spacing with the contact placeholder icons Simulator Screenshot - iPhone 14 Pro - 2023-09-06 at 15 38 16
3. I noticed that the location picker comes back up after choosing "Detect my location" Simulator Screenshot - iPhone 14 Pro - 2023-09-06 at 15 38 31

FiveCalls/FiveCalls/LocationSheet.swift
```diff
@@ -0,0 +1,141 @@
...
+//  Copyright © 2023 5calls. All rights reserved.
+//
+
+import SwiftUI
+import CoreLocation
+
+struct LocationSheet: View {
+    @Environment(\.dismiss) var dismiss
+    
+    let location: NewUserLocation?
```
cselin, line 15: It doesn't look like this is used. Did you plan to populate it when already chosen?

cselin:
> Inspired to refactor what the issuesmanager knows about, moving it from updating the app delegate itself to calling a function on an arbitrary delegate. Not even sure we need an issues manager if we do things this way... maybe just one operation executor that knows how to run all the various operations?

Leveraging unidirectional data flow is other a store could help remove the need for most of these managers. We'd just call update functions on the store and observers would be notified or any changes.

FiveCalls/FiveCalls/AdaptiveSheet.swift
```diff
@@ -0,0 +1,152 @@
...
+//
+//  Created by Nick O'Neill on 8/13/23.
+//  Copyright © 2023 5calls. All rights reserved.
+//
+
+import SwiftUI
+
+// adapted from this stack overflow answer so we can support half-screen sheets in iOS 15
+// https://stackoverflow.com/a/67994666/9137
+// replacable with presentationDetents when min iOS version is 16
```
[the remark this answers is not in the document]
nickoneill: I went back and forth on this a bit, but I think we can do iOS 16+ for simplicity. There really aren't *that* many users on iOS 15 and new ones can always use the website for core functionality. I'll bump the requirement and remove these extras.

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -147,3 +165,32 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return ProcessInfo.processInfo.environment.keys.contains("XCInjectBundleInto")
     }
 }
+
+extension AppDelegate: AppStateDelegate {
+    func setIssues(issues: [Issue]) {
+        DispatchQueue.main.async {
```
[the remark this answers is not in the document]
nickoneill: I kept the AppDelegate so we could "dual boot" the UIKit and SwiftUI interfaces during development, we'll definitely move to a scene when it's done. I haven't really grokked *why* the actor stuff was introduced yet... is it just some sugar around calls to `main.async` like this or is there another purpose that I'm missing?

FiveCalls/FiveCalls/NewUserLocation.swift
```diff
@@ -0,0 +1,113 @@
...
+    var locationType: LocationType
+    var locationValue: String
+    var locationDisplay: String
+    
+    init(address: String, display: String? = nil, gotDisplay: ((NewUserLocation) -> Void)? = nil) {
+        locationType = .address
+        locationValue = address
+        
+        // we can cache the display name when we've already fetched it, otherwise we'll go get it
+        if display != nil {
```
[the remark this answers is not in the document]
nickoneill: or even just `if let display {` here :)

FiveCalls/FiveCalls/Contact.swift
```diff
@@ -99,3 +99,38 @@ extension Contact {
...
+                    Contact(id: "1235", area: area, name: area, party: area, phone: "", photoURL: nil, fieldOffices: [])
+                ]
+        default:
+            return [Contact(id: "1234", area: area, name: area, party: area, phone: "", photoURL: nil, fieldOffices: [])]
+        }
+    }
+}
+
+// AreaToNiceString converts an area name to a generic office name that can be used in the interface
+func AreaToNiceString(area: String) -> String {
```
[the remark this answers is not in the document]
nickoneill: Good idea, I'll take a crack at this in a separate pr

FiveCalls/FiveCalls/NewUserLocation.swift
```diff
@@ -0,0 +1,113 @@
...
+        locationDisplay = "..."
+        
+        if display != nil {
+            locationDisplay = display!
+            gotDisplay?(self)
+            return
+        }
+        
+        getLocationInfo(from: location) { locationInfo in
+            self.locationDisplay = locationInfo["displayName"] as? String ?? "Unknown location"
```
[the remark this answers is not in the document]
nickoneill: 👍

FiveCalls/FiveCalls/NewUserLocation.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+extension NewUserLocation : CustomStringConvertible {
+    var description: String {
+        return "type: \(locationType.rawValue) value: \(locationValue) | display: \(locationDisplay)"
+    }
+}
+
+//extension UserLocation {
```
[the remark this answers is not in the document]
nickoneill: Removed for now, we still need to do this to the scripts but it doesn't belong in this file.

FiveCalls/FiveCalls/IssueDetail.swift
```diff
@@ -0,0 +1,50 @@
...
+                Text(issue.markdownIssueReason)
+                    .padding(.bottom, 16)
+                Text("Relevant representatives for this issue:")
+                    .font(.caption)
+                    .foregroundColor(.secondary)
+                    .padding(.bottom, 2)
+                    .padding(.leading, 6)
+                VStack(spacing: 0) {
+                    ForEach(0..<3, id: \.self) { count in
+                        ContactListItem(contact: .housePreviewContact)
```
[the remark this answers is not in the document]
nickoneill: Yep, we can update this now. This was me wanting some data for working on the design before I had that plumbed in

nickoneill:
@cselin how do the data update functions get into all the components that need to call them in a unidirectional data flow world? Put the update functions on `AppState` and then access with `@EnvironmentObject var appState: AppState` in each component that needs to update the state?

FiveCalls/FiveCalls/AppDelegate.swift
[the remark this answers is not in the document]
cselin: 👍🏽 agree with dual boot. This [section](https://www.avanderlee.com/swift/mainactor-dispatch-main-thread/#when-should-i-use-the-mainactor-attribute) spells out well an advantage to using `@MainActor`, especially with async/await. It'll be a bit more work to convert fully to async/await so we'll have to rely on operator a little longer.

FiveCalls/FiveCalls/Colors.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by Nick O'Neill on 8/25/23.
+//  Copyright © 2023 5calls. All rights reserved.
+//
+
+import SwiftUI
+
+extension Color {
+    static let FiveCallsLightBG = Color(red: 0.85, green: 0.85, blue: 0.85)
+    static let FiveCallsDarkBlue = Color("darkBlue")
```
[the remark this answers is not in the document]
cselin: I'm actually not convinced this will work, but see my comment [here](https://github.com/5calls/ios/pull/337#discussion_r1327418576) for a better alternative.

FiveCalls/FiveCalls/NewUserLocation.swift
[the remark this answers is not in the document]
cselin: TIL 😀

cselin:
> @cselin how do the data update functions get into all the components that need to call them in a unidirectional data flow world? Put the update functions on `AppState` and then access with `@EnvironmentObject var appState: AppState` in each component that needs to update the state?

More or less, yes. A full blown store with middleware and the like might be overkill for this size app. If `AppState` blew up too much it could be factored apart later. Ideally there'd only be a single point of access for updating state so we wouldn't need to access them globally, it would just be part of the flow. For example in `Dashboard`, we would send a `fetchIssues` action on the store (or AppState in our case) which would update the state directly so we wouldn't have to rely on callbacks or passing around the app delegate.

FiveCalls/FiveCalls/Colors.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by Nick O'Neill on 8/25/23.
+//  Copyright © 2023 5calls. All rights reserved.
+//
+
+import SwiftUI
+
+extension Color {
+    static let FiveCallsLightBG = Color(red: 0.85, green: 0.85, blue: 0.85)
+    static let FiveCallsDarkBlue = Color(R.color.darkBlue()!)
```
cselin, line 13: Feel free to cherry pick my change [here](https://github.com/5calls/ios/pull/337/commits/6a1262ed8b0d7d55358c76b689e80f78081d797e) to allow removing this extension all together.

nickoneill: Turns out Xcode 15 has a new setting that will generate color extensions for you! I prefixed the color asset names so it's clear which ones are nonstandard. I tried the same thing with the new xcstrings support, but I'm not sure I like how automatic it is... I prefer seeing the autocompletion for existing strings like R.swift shows.

cselin: Nice! I see you updated some colors still using R, would we be able to convert all the colors off of R, or is there still a use case?

nickoneill: I don't plan on making improvements to the UIKit code where we can avoid it, we can get rid of it during the dependency cleanup phase. SwiftUI stuff should all be the new Xcode color extensions though, I'll take a look if I missed anything.

nickoneill:
@cselin take a look at my attempt at unidirectional data flow? I figured that using `@MainActor` would absolve me of having to `DispatchQueue.main.async` because that could be inferred but apparently not? Or am I holding it wrong. I am not grokking the point of Actors if I still have to sprinkle dispatch queues where I know I'll be updating UI components.

FiveCalls/FiveCalls/AppState.swift
```diff
@@ -49,9 +49,26 @@ class AppState: ObservableObject {
...
-    func setIssues(issues: [Issue])
-    func setContacts(contacts: [Contact])
-    func setFetchingContacts(fetching: Bool)
-    func setLocation(location: NewUserLocation)
+extension AppState {
+    @MainActor
+    func setLocation(newLocation: NewUserLocation) {
+        location = newLocation
+        
+        Operator().fetchContacts(location: newLocation, fetching: self.setFetchingContacts, setContacts: self.setContacts)
```
cselin, line 57: So we wouldn't want to use `MainActor` for this whole method since I think it would then run the operation on the main thread, which we don't want.

cselin: I figured we could indicate the whole `AppState` class as `MainActor` and wouldn't have to repeat it for each function below, however, for this one or at least the fetch, we'd want to disqualify.

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -25,7 +25,7 @@ struct Dashboard: View {
                             showLocationSheet.toggle()
                         }
                         .sheet(isPresented: $showLocationSheet) {
-                            LocationSheet(location: appState.location, delegate: (UIApplication.shared.delegate as! AppDelegate))
+                            LocationSheet(location: appState.location, setLocation: appState.setLocation)
```
cselin, line 28: Instead of passing this tightly coupled completion function, would it make more sense to just include the app state as an environment object? That way, if the mechanisms change for what we do in the location sheet, we can just adjust what we call from there.

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -48,26 +48,11 @@ struct Dashboard: View {
...
-                    op.fetchContacts(location: location, delegate: (UIApplication.shared.delegate as! AppDelegate)) { result in
-                        if case let .serverError(error) = result {
-                            print("contacts error: \(error)")
-                        }
-                        if case .offline = result {
-                            print("contacts offline")
-                        }
-        
-                    }
+                    op.fetchContacts(location: location, fetching: appState.setFetchingContacts, setContacts: appState.setContacts)
```
cselin, line 55: Likewise, just allowing access to the app state from these fetch operations, we could avoid passing in the handlers.

FiveCalls/FiveCalls/LocationSheet.swift
```diff
@@ -13,7 +13,7 @@ struct LocationSheet: View {
     @Environment(\.dismiss) var dismiss
     
     let location: NewUserLocation?
-    let delegate: AppStateDelegate?
+    let setLocation: @MainActor (NewUserLocation) -> () // really? you have to specify mainactor here too?
```
cselin, line 16: I too admit I don't entirely fully grok `MainActor` yet, even after reading up on it a bit more. It sounds magical, like you just apply it and everything within scope runs on the main queue.

FiveCalls/FiveCalls/LocationSheet.swift
```diff
@@ -106,7 +106,9 @@ struct LocationSheet: View {
     
     func locationSearch() {
         _ = NewUserLocation(address: locationText) { loc in
-            delegate?.setLocation(location: loc)
+            DispatchQueue.main.async {
+                setLocation(loc)
```
cselin, line 110: Hmm, I would have thought declaring `MainActor` above would have avoided needing this. Maybe putting the entire `AppState` object within `MainActor` would do the trick?

FiveCalls/FiveCalls/Operator.swift
```diff
@@ -23,39 +23,43 @@ class Operator {
...
 
-    func fetchIssues(delegate: AppStateDelegate, completion: @escaping (OperationResult) -> Void) {
+    func fetchIssues(setIssues: @MainActor @escaping ([Issue]) -> ()) {
         let operation = FetchIssuesOperation()
         operation.completionBlock = { [weak operation] in
             if let issues = operation?.issuesList {
                 DispatchQueue.main.async {
-                    delegate.setIssues(issues: issues)
-                    completion(.success)
+                    setIssues(issues)
```
cselin, line 31: In a unidirectional world of the type I'm used to this would be more like `appStore.send(.setIssues(issues))` where the store contains the current `appState` and has a reducer that handles the actions wrt the state.

FiveCalls/FiveCalls/Operator.swift
```diff
@@ -23,39 +23,43 @@ class Operator {
...
-                DispatchQueue.main.async {
-                    print(error.localizedDescription)
-                    
-                    if error.isOfflineError() {
-                        completion(.offline)
-                    } else {
-                        completion(.serverError(error))
-                    }
-                }
+                // TODO(nickoneill): do we need to return an error? does it make more sense to have a function that sets an error in app state? or make this async with a completion so state in the component can handle the error?
```
cselin, line 36: Completions seem to be getting phased out in lieu of async/await and setting error in the app state would be the SwiftUI-y way to do it.

cselin (commented):
I think we need to think a bit more on how to properly use `MainActor` since it *should* help avoid additional dispatch calls.

For unidirectional, I think keeping the methods on the app delegate for now is more in line with sending actions on an App Store and is less tightly coupled than migrating them to handler closures.

I'll attempt a quick unidirectional approach commit to demonstrate and see if its worth going ahead with that sort of thing for now.

FiveCalls/FiveCalls/Operator.swift
cselin: @nickoneill seem my rough approach here: https://github.com/5calls/ios/pull/338

FiveCalls/FiveCalls/AppState.swift
nickoneill: I guess I was continuing the assumption that the operation execution runs on some background thread regardless of the context it was called on. I don't actually know how this interacts with `@MainActor`!

FiveCalls/FiveCalls/Dashboard.swift
nickoneill: Yeah, that does make more sense.

FiveCalls/FiveCalls/LocationSheet.swift
nickoneill: I read a bit more about actors in general, which I think I get, but I am still baffled by how you should structure `@MainActor` so you don't end up with even *more* main-handling code.

FiveCalls/FiveCalls/Operator.swift
nickoneill: resolved with the unidirectional data flow

FiveCalls/FiveCalls/Operator.swift
nickoneill: more of this approach is in this branch now

FiveCalls/FiveCalls/Dashboard.swift
nickoneill: resolved with proper unidirectional data flow

nickoneill:
@cselin I merged your approach back into this branch, removing _a tiny bit_ of the generics in the reducer for simplicity and because I wanted to dispatch a contact fetch when the location was changed. I added some error setting as well but it's not reflected anywhere in the views yet.

If this approach seems right to extend, I think it's time to get this merged back to main so we can better manage branches with incremental improvements going forward.

nickoneill:
Oh and I refactored actions into an enum, the types are a bit more concisely named and we can rely on switches being exhaustive to ensure all the cases are covered.

FiveCalls/FiveCalls/Store.swift
```diff
@@ -0,0 +1,59 @@
...
+        switch action {
+        case let action as SetFetchingContactsAction:
+            state.fetchingContacts = action.fetching
+        case let action as SetIssuesAction:
+            state.issues = action.issues
+        case let action as SetContactsAction:
+            state.contacts = action.contacts
+        case let action as SetLocationAction:
+            state.location = action.location
+            self.dispatch(action: FetchContactsAction(location: action.location))
```
cselin: I *think* this call to dispatch belongs in the middleware

cselin (commented):
Looks great, just one minor suggestion to move a dispatching an action from the reducer to middleware

FiveCalls/FiveCalls/Actions.swift
```diff
@@ -7,35 +7,13 @@
...
-    let location: NewUserLocation
-}
-
-struct SetContactsAction: Action {
-    let contacts: [Contact]
-}
-
-struct SetLocationAction: Action {
-    let location: NewUserLocation
+enum Action {
```
cselin, line 10: This will work as long as we only have a single action type, which may be the case with an app this small. I think it makes sense at this point for simplicity 👍🏽

FiveCalls/FiveCalls.xcodeproj/project.pbxproj
```diff
@@ -557,6 +604,7 @@
 				B512506921E6F762007D32DF /* FetchContactsOperation.swift */,
 				9C46B1F91E46EA58009B2C11 /* ReportOutcomeOperation.swift */,
 				A0D474B3202A22B90012BC94 /* FetchUserStatsOperation.swift */,
+				D07829742A7B32A8001EE5DB /* Operator.swift */,
```
cselin, line 607: I think we can get rid of this now, right?

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -0,0 +1,80 @@
...
+//
+
+import SwiftUI
+
+struct Dashboard: View {
+    @EnvironmentObject var store: Store
+
+    @State var showLocationSheet = false
+
+    let op = Operator()
```
cselin, line 16: can remove

FiveCalls/FiveCalls/Appearance.swift
```diff
@@ -12,10 +12,15 @@ enum Appearance {
     static func setup() {
         let pageControlAppearance = UIPageControl.appearance()
         pageControlAppearance.pageIndicatorTintColor = R.color.lightBlue()
-        pageControlAppearance.currentPageIndicatorTintColor = R.color.darkBlue()
+        pageControlAppearance.currentPageIndicatorTintColor = R.color.fivecallsDarkBlue()
```
cselin, line 15: This doesn't seem to be working for me in SwiftUI branch, page indicators are missing.

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -7,13 +7,17 @@
...
 import OneSignal
 import Firebase
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
-
     var window: UIWindow?
+    var navController: CustomNavigationController!
+    
+    let USE_NEW_SWIFTUI_INTERFACE = true
```
cselin, line 19: should set this to `false` before committing 😉

FiveCalls/FiveCalls/5CallsAPIMiddleware.swift
```diff
@@ -0,0 +1,72 @@
+//
+ //  Middlewares.swift
+ //  FiveCalls
+ //
+ //  Created by Christopher Selin on 9/22/23.
+ //  Copyright © 2023 5calls. All rights reserved.
+ //
+ import Foundation
+
+ func appMiddleware() -> Middleware<AppState> {
```
cselin, line 10: nit: should match filename?

FiveCalls/FiveCalls/AppState.swift
```diff
@@ -0,0 +1,51 @@
...
+            guard let location = self.location else { return }
+            let defaults = UserDefaults.standard
+            defaults.set(location.locationType.rawValue, forKey: UserDefaultsKey.locationType.rawValue)
+            defaults.set(location.locationValue, forKey: UserDefaultsKey.locationValue.rawValue)
+            defaults.set(location.locationDisplay, forKey: UserDefaultsKey.locationDisplay.rawValue)
+            Logger().info("saved cached location as \(location)")
+        }
+    }
+    @Published var fetchingContacts = false
+    @Published var issueLoadingError: Error? = nil
```
cselin, line 27: add TODO to integrate these errors?

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -0,0 +1,80 @@
...
+                            showLocationSheet.toggle()
+                        }
+                        .sheet(isPresented: $showLocationSheet) {
+                            LocationSheet()
+                                .presentationDetents([.medium])
+                                .presentationDragIndicator(.visible)
+                                .padding(.top, 40)
+                            Spacer()
+                        }
+                    Text("What’s important to you?")
```
cselin, line 34: localize

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -0,0 +1,80 @@
...
+                        .font(.system(size: 20))
+                        .fontWeight(.semibold)
+                    ForEach(store.state.issues) { issue in
+                        NavigationLink(value: issue) {
+                            IssueListItem(issue: issue, contacts: store.state.contacts)
+                        }
+                    }
+                }.padding(.horizontal, 10)
+            }.navigationTitle("Issues")
```
cselin, line 43: localize

FiveCalls/FiveCalls/Issue.swift
```diff
@@ -26,10 +26,64 @@ struct Issue : Decodable {
...
+
+        return sortedContacts
+    }
+    
+    // sortedContactAreas takes a list of contact areas and orders them in our preferred order,
+    // we should always order them properly on the server but let's do this to be sure
+    func sortedContactAreas(areas: [String]) -> [String] {
+        var contactAreas: [String] = []
+        
+        for area in ["StateLower", "StateUpper", "US House", "US Senate", "Governor", "AttorneyGeneral", "SecretaryOfState"] {
```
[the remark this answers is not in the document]
cselin: maybe add a TODO for this?

FiveCalls/FiveCalls/IssueContactDetail.swift
```diff
@@ -0,0 +1,68 @@
...
+                .padding(.bottom)
+            Text(issue.markdownIssueScript)
+                .padding(.horizontal)
+            PrimaryButton(title: nextButtonTitle(), systemImageName: "megaphone.fill")
+                .padding()
+            Spacer()
+        }
+    }
+    
+    func nextButtonTitle() -> String {
```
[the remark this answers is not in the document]
cselin: still applies

FiveCalls/FiveCalls/LocationSheet.swift
```diff
@@ -0,0 +1,141 @@
...
+                Text("Use an address, zip code or zip + 4")
+                    .font(.footnote)
+                    .foregroundColor(.secondary)
+                    .padding(.leading, 35)
+            }
+            .padding(.bottom)
+            HStack(alignment: .top) {
+                Text("Or")
+                    .font(.system(.title3))
+//                    .fontWeight: .regular)
```
cselin, line 67: remove?

FiveCalls/FiveCalls/NewUserLocation.swift
```diff
@@ -0,0 +1,97 @@
...
+                return nil
+            }
+        }
+    }
+        
+    var locationType: LocationType
+    var locationValue: String
+    var locationDisplay: String
+    
+    init(address: String, display: String? = nil, gotDisplay: ((NewUserLocation) -> Void)? = nil) {
```
cselin, line 36: no need for changes now, but perhaps we can integrate this with our unidirectional store to observe changes instead of requiring a callback closure? Using Combine to observe changes to `locationDisplay` on the current location object in the store if that's the property we care most about.

FiveCalls/FiveCalls/NewUserLocation.swift
```diff
@@ -0,0 +1,97 @@
...
+            return
+        }
+        
+        getLocationInfo(from: location) { locationInfo in
+            self.locationDisplay = locationInfo["displayName"] as? String ?? R.string.localizable.unknownLocation()
+            gotDisplay?(self)
+        }
+    }
+    
+    private func getLocationInfo(from location: CLLocation, completion: @escaping (([String: Any]) -> ())) {
```
cselin, line 77: another nice to have: this is a good candidate for converting to async/await

FiveCalls/FiveCalls/Operator.swift
```diff
@@ -0,0 +1,97 @@
...
+
+enum OperationResult {
+    case success
+    case serverError(Error)
+    case offline
+}
+
+// Operator is responsible for running any of our operation classes and managing the results. Since we store the state in AppState
+// rather than some other manager instance, we don't need some intermediary state holder
+class Operator {
```
cselin, line 19: we can remove this file now, right?

cselin (approved):
as long as we default SwiftUI off, I think this is in good shape to merge

FiveCalls/FiveCalls/Store.swift
nickoneill: oh yeah of course, moved.

FiveCalls/FiveCalls.xcodeproj/project.pbxproj
nickoneill: Yep, gone

FiveCalls/FiveCalls/Appearance.swift
nickoneill: Probably ok for now, added a bug to #331

FiveCalls/FiveCalls/5CallsAPIMiddleware.swift
nickoneill: done, back to just middleware for now

FiveCalls/FiveCalls/AppState.swift
nickoneill: 👍

FiveCalls/FiveCalls/Dashboard.swift
nickoneill: There's a number of missing localizations, added an item to review them all in #331

FiveCalls/FiveCalls/NewUserLocation.swift
nickoneill: Definitely, adding some todos

PR #330, opened 2023-07-24
Fix build cycle error for Xcode 17

Abizern:
When building with Xcode 17 there is a error caused by a build cycle being detected.

Turns out this is an issue with the Crashlytics script and fixed with the suggestion here: https://github.com/firebase/firebase-ios-sdk/issues/11471


tomassliz:
@Abizern I think you got the wrong version of Xcode. You mean Xcode 15.
Anyway, in the linked discussion they say that this issue was already solved in some beta. But not for me. I'm using the latest Xcode 15.0 (15A240d). What's your experience, please?

PR #332, opened 2023-09-06
Remove broken open source libraries cell

cselin:
|Before|After|
|-|-|
|Simulator Screenshot - iPhone 14 Pro - 2023-09-06 at 11 41 31|Simulator Screenshot - iPhone 14 Pro - 2023-09-06 at 11 40 30|

Removes "Credits" header and non-functional "Open Source Libraries" cell, preserving version footer.

It looks like CPDAcknowledgements is no longer being actively supported and there weren't any contributions being passed into it anyway.

### DIFF
--- a/FiveCalls/FiveCalls/AboutHtmlViewController.swift
+++ b/FiveCalls/FiveCalls/AboutHtmlViewController.swift
@@ -8,7 +8,6 @@
 
 import Foundation
 import UIKit
-//import CPDAcknowledgements
 import WebKit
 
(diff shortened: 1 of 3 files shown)


nickoneill (approved):
Thanks @cselin! I made #333 to track reintegration of something like this for the SwiftUI work we're doing.

Apparently I have to approve first time contributors to run workflows? but I don't see a place to do that, hence bitrise failing. We can merge without for now.

Issue #216, opened 2018-08-06
Increase the contrast of the text on the "Enable Reminder" screen


foozmeat:
I find it hard to read on my phone. I think using the regular blue color here would be fine. 

![img_aaca91d653c2-1](https://user-images.githubusercontent.com/419456/43739831-3d744066-997e-11e8-9406-4a81c1630e25.jpeg)

subdigital:
Agree it’s hard to read but we should avoid using blue color for things that aren’t interactive.

foozmeat:
fair enough

nickoneill:
What about good old #000000

subdigital:
I think lower contrast is better, but perhaps not as much as it is now.

heathermh:
Fixed by https://github.com/5calls/ios/pull/374

PR #334, opened 2023-09-06
Increase contrast of enable reminder text

cselin:
Per [this issue](https://github.com/5calls/ios/issues/216), it's hard to distinguish the text from the background prior to enabling reminders.

I'm leaning towards Medium Dark Gray, although Dark Gray doesn't look bad either.

|Before (Light Gray)|After (Medium Dark Gray)|Medium|Dark Gray|
|-|-|-|-|
|before|medium-dark-gray|medium|Dark-gray|


nickoneill:
Still feels like Medium is hard to distinguish, I think Dark Gray is the right way to go

nickoneill (approved):

PR #338, opened 2023-09-22
SwiftUI Unidirectional

cselin:
Rough migration to simple Redux unidirectional flow, removing the need for tightly coupled completions, delegates and even the Operator.


cselin:
@nickoneill take a look at this rough approach.

PR #339, opened 2023-10-03
Connect contacts through script pages, hook up a done page at the end

nickoneill:
Plumbed the real contacts through the script pages, added a placeholder done for now


FiveCalls/FiveCalls/IssueContactDetail.swift
```diff
@@ -50,19 +50,38 @@ struct IssueContactDetail: View {
...
+    func currentContact() -> Contact {
+        return remainingContacts.first!
+    }
+    
+    func nextContacts() -> [Contact] {
+        return Array(remainingContacts.dropFirst())
+    }
+    
     func nextButtonTitle() -> String {
         return "Next contact"
```
cselin, line 79: localize

FiveCalls/FiveCalls/IssueContactDetail.swift
```diff
@@ -50,19 +50,38 @@ struct IssueContactDetail: View {
...
     
+    func currentContact() -> Contact {
+        return remainingContacts.first!
+    }
+    
+    func nextContacts() -> [Contact] {
+        return Array(remainingContacts.dropFirst())
+    }
+    
     func nextButtonTitle() -> String {
```
cselin, line 78: this can probably just be replaced with localized string inline

FiveCalls/FiveCalls/IssueDetail.swift
```diff
@@ -26,15 +28,15 @@ struct IssueDetail: View {
...
+                        ContactListItem(contact: contact.element)
+                        if contact.number < 2 { Divider().padding(0) } else { EmptyView() }
                     }
                 }.background {
                     RoundedRectangle(cornerRadius: 10)
                         .foregroundColor(Color(red: 0.85, green: 0.85, blue: 0.85))
                 }.padding(.bottom, 16)
-                NavigationLink(destination: IssueContactDetail(issue: issue, contact: .housePreviewContact)) {
+                NavigationLink(destination: IssueContactDetail(issue: issue, remainingContacts: issue.contactsForIssue(allContacts: store.state.contacts))) {
                     PrimaryButton(title: "See your script", systemImageName: "megaphone.fill")
```
cselin, line 40: localize

cselin (approved):

FiveCalls/FiveCalls/Contact.swift
```diff
@@ -98,6 +98,27 @@ extension Contact {
         let scriptWithContactName = regex.stringByReplacingMatches(in: script, options: [], range: fullRange, withTemplate: template)
         return scriptWithContactName
     }
+    
+    // this has some overlap with other area -> string conversions but I haven't thought about it long enough to combine them
+    func officeDescription() -> String {
```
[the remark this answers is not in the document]
nickoneill: 👍

FiveCalls/FiveCalls/IssueContactDetail.swift
```diff
@@ -50,19 +50,38 @@ struct IssueContactDetail: View {
...
+                        PrimaryButton(title: "Done calling",
+                                      systemImageName: "megaphone.fill")
+                            .padding()
+                    }
+            }
             Spacer()
         }
     }
     
+    func currentContact() -> Contact {
```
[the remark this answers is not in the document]
nickoneill: No, they fit better as properties, changed.

nickoneill:
@cselin addressed your comments and added a router + better applied the navigation stack concept. I'm not sure I like the navigation models, I was trying to do the same with just tuples but I suppose those wouldn't be hashable (and I couldn't figure out the right type signature). Still, I don't dislike the concept of having all the nav destinations at the root stack.

FiveCalls/FiveCalls/IssueDetail.swift
```diff
@@ -38,11 +38,12 @@ struct IssueDetail: View {
                     RoundedRectangle(cornerRadius: 10)
                         .foregroundColor(Color(red: 0.85, green: 0.85, blue: 0.85))
                 }.padding(.bottom, 16)
-                NavigationLink(destination: IssueContactDetail(issue: issue, remainingContacts: issue.contactsForIssue(allContacts: store.state.contacts))) {
-                    PrimaryButton(title: "See your script", systemImageName: "megaphone.fill")
+                NavigationLink(value: IssueDetailNavModel(issue: issue, contacts: issue.contactsForIssue(allContacts: store.state.contacts))) {
```
cselin, line 41: nit: I wonder if we need to pass the main app store into this view, it looks like we're only using it to access the contacts, but not performing any actions. We could probably just pass in the contacts to this view

FiveCalls/FiveCalls/IssueDone.swift
```diff
@@ -9,13 +9,17 @@
...
+    
     let issue: Issue
     
     var body: some View {
         ScrollView {
             VStack {
                 Text("Issue \(issue.name) done page")
-                Button(action: {}, label: {
+                Button(action: {
+                    router.backToRoot()
```
cselin, line 21: nit: this could be done via the unidirectional store instead if we wanted to keep everything in that flow. We could add an action `backToRoot` (handled in middleware) that could be sent from anywhere that has access to the store without having to expose the routing details.

FiveCalls/FiveCalls/Router.swift
```diff
@@ -0,0 +1,53 @@
...
+    func back() {
+        path.removeLast()
+    }
+    
+    func backToRoot() {
+        path = NavigationPath()
+    }
+}
+
+struct IssueDetailNavModel {
```
cselin, line 23: nit: these probably belong in their own file, or maybe in the view files that use them?

FiveCalls/FiveCalls/Router.swift
```diff
@@ -0,0 +1,53 @@
...
+//  File.swift
+//  FiveCalls
+//
+//  Created by Nick O'Neill on 10/8/23.
+//  Copyright © 2023 5calls. All rights reserved.
+//
+
+import SwiftUI
+
+class Router: ObservableObject {
```
cselin, line 11: love this!

cselin:
> @cselin addressed your comments and added a router + better applied the navigation stack concept. I'm not sure I like the navigation models, I was trying to do the same with just tuples but I suppose those wouldn't be hashable (and I couldn't figure out the right type signature). Still, I don't dislike the concept of having all the nav destinations at the root stack.

Love the router. wrt the nav models, we did something similar at my last company, though we called them view state and embedded them in the views. They were similar in concept to view models that represented a subset of the state they were derived from that we just needed in a respective view. This was all pre-navigation stack so it's interesting to see what must be done to keep it hashable. I think what you've done here works well.

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -10,13 +10,14 @@ import SwiftUI
 
 struct Dashboard: View {
     @EnvironmentObject var store: Store
+    @EnvironmentObject var router: Router
```
cselin, line 13: wondering if this particular router belongs directly on this view only, rather than as an environmental object. Since popping to root will only ever go back to the dashboard it seems like it could be restricted to here. Just a thought.

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -39,7 +39,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         window = UIWindow()
         if USE_NEW_SWIFTUI_INTERFACE {
             let store = Store(state: AppState(), middlewares: [appMiddleware()])
-            window?.rootViewController = UIHostingController(rootView: Dashboard().environmentObject(store))
+            let router = Router()
```
cselin, line 42: see my comments elsewhere about 1) restricting this dashboard specific router to the dashboard and subviews and 2) hiding this behind the store so consumers don't need to know the internals about how routing works.

cselin (approved):
Looks good. I proposed some ideas about adding a bit more separation for the new router for consideration.

FiveCalls/FiveCalls/IssueDetail.swift
nickoneill: I don't quite understand; issue detail only gets an issue and an array of the remaining contacts, they are filtered by the `contactsForIssue` method which matches on area.

FiveCalls/FiveCalls/Dashboard.swift
nickoneill: Are you suggesting this in conjunction with moving router actions into middleware? Intermediate screens currently need to know about the router as well so they can pop a single item off the path to go back.

FiveCalls/FiveCalls/Router.swift
nickoneill: 👍 will move these into the views that use that model

FiveCalls/FiveCalls/IssueDetail.swift
cselin: Right, that's what I mean, we include the main store on line 12 of this file, but if we're only using it to get the contacts from the state, we could just include pass the contacts into the initializer for this view, unless you think we'll need access to the full store in the future. Just being overly picky here if we wanted to limit access is all.

FiveCalls/FiveCalls/Dashboard.swift
cselin: Not necessarily tied to router actions in the middleware. Just thinking that this router could be an @StateObject declared and initialized here if the router is only used in the Dashboard and subviews. Arguably, the same could be said for the store since everything in it applies to the dashboard and subviews, so it's probably fine to keep both declared and initialized one level up in the app. You can disregard this thread ^

FiveCalls/FiveCalls/IssueDetail.swift
nickoneill: ok, I get it. Yeah, I'll give that a try

FiveCalls/FiveCalls/IssueDone.swift
nickoneill: I'm going to give this a shot to see how it feels but I'll do it in another PR

PR #340, opened 2023-10-04
About converted to SwiftUI

cselin:
|Before|After|
|-|-|
|Simulator Screenshot - iPhone 15 Pro - 2023-09-26 at 17 53 58|Simulator Screenshot - iPhone 15 Pro - 2023-09-29 at 21 06 01|
|Simulator Screenshot - iPhone 15 Pro - 2023-09-26 at 20 19 59|Simulator Screenshot - iPhone 15 Pro - 2023-09-26 at 20 17 43|

Notes:
- relocates the About screen to the five stars logo in top right which may not be super discoverable.
- I kept the style pretty close to the original, but if we wanted to use more modern SwiftUI defaults, we could get away from some of the custom code. For example, if we only showed chevrons on the actual link cells, we wouldn't have to do extra work in `AboutListItem`.
- I couldn't find a way to exclude items from `ShareLink` as we did with the old share sheet.


cselin:
@nickoneill please take a look now that I rebased on the latest SwiftUI changes

nickoneill (approved):
Thanks, this looks great. I think we'll tweak the style to be more swiftui-like eventually but it's great just to have it out of UIKit for now

PR #341, opened 2023-10-05
Resolve some warnings

nickoneill:
Renamed some colors that were conflicting with the new generated colors in Xcode 15, these were mostly in UIKit.

I also bumped the notification extension version to match the app... but it doesn't seem like there's a way to update it in the target editor right now (except in the build settings view where everything is listed). I could have sworn it was in "General" for previous versions of Xcode.


FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -16,7 +16,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
     var navController: CustomNavigationController!
     
-    let USE_NEW_SWIFTUI_INTERFACE = false
```
cselin, line 19: revert 😄

cselin (approved):
👍🏽 after reverting swift ui var

FiveCalls/FiveCalls/AppDelegate.swift
nickoneill: 🤦‍♂️

PR #342, opened 2023-10-05
Fix snapshot tests

cselin:


FiveCalls/FiveCallsUITests/GET-v1-reps.json
```diff
@@ -17,7 +20,7 @@
       "id": "CA-NancyPelosi",
       "name": "Nancy Pelosi",
       "phone": "202-225-4965",
-      "photoURL": "https://repimages.5calls.org/us-house/512x512/nancy-pelosi.jpg",
+      "photoURL": "https://images.5calls.org/house/256/P000197.jpg",
```
cselin, line 23: Looks like the old photo URL's are working anymore. I couldn't find a working URL for Diane Feinstein since she died a few days ago (RIP).

nickoneill: Good call, it's been a while since we used these old image urls

nickoneill (approved):
I don't think we've ever compared them as a test case (in general I don't think snapshot tests are all that useful as test cases unless you have a lot of reusable components) but they are *very* useful to generate App Store screenshots, thanks for taking a look at them

PR #343, opened 2023-10-12
Migrate Your Impact to SwiftUI

cselin:
|Before|After|
|-|-|
|Simulator Screenshot - iPhone 15 Pro - 2023-10-06 at 11 30 41|IMG_45CEF0F9AE85-1|
|IMG_2169|IMG_2171|

|Menu|
|-|
|IMG_2174|


cselin:
@nickoneill FYI: the server stats don't seem to be working for me. I fudged the numbers in the screenshots above to match.

Also, I just noticed a minor layout issue when there are zero calls and the text is centered instead of left aligned.

cselin:
Also wondering if we should show a tooltip or something on the gear menu the first few times people use the app so there's more awareness on where Your Impact and About went?

nickoneill:
Yep, it looks like stats haven't been updating since I migrated to a new callerid anonymous id recently. Add a new line in `FetchUserStatsOperation` when preparing the request:

`request.setValue(AnalyticsManager.shared.callerID, forHTTPHeaderField: "X-Caller-ID")`

Everything else looks great - do you want to try out conditionally using TipKit in iOS 17 for the gear icon? Could be a separate PR.

nickoneill (approved):

cselin:
@nickoneill I'm still not sure how to make Bitrise happy, I tried logging in with my GitHub account which now has access to this repo, but it looks like the latest check passed and when I attempt to rebuild a failed check, it no-op's 🤔

nickoneill:
This PR is still from your fork which I don't know how to automatically approve on bitrise. I added you as a collaborator on this repo, try pushing a branch directly from/to this repo next time and it should run bitrise automatically.

I agree we need a local source of at least the total number of calls (we used to store json for each call outcome but I think that's unnecessary with the server stats). We key a few prompts for reviews, donations, etc, after making a certain number of calls so having a total in the appstate would be great.

All good on tipkit in another PR.

PR #345, opened 2023-10-23
Migrate Welcome to SwiftUI and total stats to store

cselin:
https://github.com/5calls/ios/assets/810263/178768c0-c0ae-4b2c-8c62-5d4a726c3deb


cselin:
@nickoneill here's what I have so far. I'm thinking maybe we should at a TTL so that stats can be updated in session if a certain amount of time has passed and it is reasonable that they may have made a call.

cselin:
> @nickoneill here's what I have so far. I'm thinking maybe we should at a TTL so that stats can be updated in session if a certain amount of time has passed and it is reasonable that they may have made a call.

@nickoneill let me know what you think about a TLL or not

FiveCalls/FiveCalls/Localizable.strings
```diff
@@ -130,3 +130,12 @@
 "called-multiple"                     = "%d times";
 "called-single"                       = "%d time";
 "subheading-message"                  = "You're making a difference! Calling is the most effective way to influence your representatives, and here's what you've achieved:";
+
+// welcome
+"welcome-page-1-title"                = "MAKE YOUR VOICE HEARD";
```
nickoneill, line 135: prefer capitalizing these in the views vs in the translation?

FiveCalls/FiveCalls/AppState.swift
```diff
@@ -11,6 +11,7 @@ import CoreLocation
 import os
 
 class AppState: ObservableObject, ReduxState {
+    @Published var numberOfCalls: Int = 0
```
nickoneill, line 14: A bit ambiguously named, since we count calls that you've made, calls on a particular issue, etc. `globalCallCount`?

nickoneill (approved):
Couple small comments, but looks great. I think we should tackle making the welcome screen look more like a modern iOS welcome screen before release but this'll make things a hell of a lot easier to do that.

I _don't_ think we need a TTL on this number though. We fetch it again when we get to the "done" page, along with the call count for that topic (redesign for this page is a PR I'm working on now) and we can update it in the state then. Anyone using the app regularly will hit a "done" page and it'll get updated that way, and anyone not using the app regularly will presumably get it fresh after the app is killed in the background between uses.

Sorry for being largely absent for iOS code stuff this week, we had > 70,000 calls made for the current Israel/palestine issue (largely through the iOS app!) and it was a good opportunity to see how the API handled the load and make some performance improvements there.

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -17,7 +17,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
     var navController: CustomNavigationController!
     
-    let USE_NEW_SWIFTUI_INTERFACE = false
+    let USE_NEW_SWIFTUI_INTERFACE = true
```
cselin, line 20: boop!

cselin:
> Sorry for being largely absent for iOS code stuff this week, we had > 70,000 calls made for the current Israel/palestine issue (largely through the iOS app!) and it was a good opportunity to see how the API handled the load and make some performance improvements there.

That's great news! Let me know if you need any help with the performance stuff.

PR #346, opened 2023-10-24
Add popover tip to settings

cselin:
https://github.com/5calls/ios/assets/810263/0045eed1-684a-436a-b415-48dbf4a42399

Note: I've since fixed the type "resides" -> "reside" since recording the video above ^

In my testing, I've only ever seen the tip twice, so it could be double counting, I've noticed it can be a bit buggy if you navigate away from the dashboard during tip presentation. I also noticed that tapping X on the tip seems to prevent it from being shown again (at least immediately). I considered hooking up tapping on settings to hiding it indefinitely as well, but it takes quite a bit more work due to it only being available in iOS 17+


cselin:
PTAL @nickoneill

nickoneill (approved):
Looks great!

PR #348, opened 2023-10-30
Updating Rswift to latest release at https://github.com/mac-cain13/R.swift

heathermh:
While helping me with my first build of the project, @nickoneill realized we were pointing to branch at https://github.com/mac-cain13/R.swift.Library.

Current recommendation is to use https://github.com/mac-cain13/R.swift instead, so I switched over to latest version of that. I wasn't able to get the older versions building anyway.

I had to pick carefully through commits, but I think this is stable.


heathermh:
@nickoneill I can go ahead and add GoogleService-Info.plist if you want. I didn't add it before / assumed you emailed it to me for reasons.

nickoneill:
I would prefer to leave it out of the repo. I think the build issue here was that it was linked in the project and the build step is looking for it. If you omit it from the project but leave it in the top level `FiveCalls` directory then the builds are happy locally and the CI is happy as well.

nickoneill:
I think there are one or two more steps to migrate this to a modern R.Swift install, [see items 3 and 4 here](https://github.com/mac-cain13/R.swift#xcode-project---spm), which would replace our old-style build phase run script with a build tool invocation. If we're updating to the modern way to integrate R.Swift, we should definitely use this modern code generation step too.

heathermh:
> I think there are one or two more steps to migrate this to a modern R.Swift install, [see items 3 and 4 here](https://github.com/mac-cain13/R.swift#xcode-project---spm), which would replace our old-style build phase run script with a build tool invocation. If we're updating to the modern way to integrate R.Swift, we should definitely use this modern code generation step too.

This was good. I updated these settings and it seems a little smoother now.

heathermh:
> I would prefer to leave it out of the repo. I think the build issue here was that it was linked in the project and the build step is looking for it. If you omit it from the project but leave it in the top level `FiveCalls` directory then the builds are happy locally and the CI is happy as well.

If I pull it from the project, I get an error on the Run Script Crashlytics step:
error: Could not get GOOGLE_APP_ID in Google Services file from build environment

Are you able to modify that step so it will work?

nickoneill:
Oh actually it looks like it is already in the project under the models group (weird place for it). It's looking for a file at `FiveCalls/GoogleService-Info.plist` so as long as you place it there, crashlytics should be able to find it.

The CI machine is reporting a different error (it will add the plist file itself) relating to "enabling" the new R.Swift helper tool?

heathermh:
> The CI machine is reporting a different error (it will add the plist file itself) relating to "enabling" the new R.Swift helper tool?

This link explains what to add to the CI. I think that's above my pay grade? Please take a look. https://github.com/mac-cain13/R.swift/blob/main/Documentation/Migration.md#configuring-continuous-integration-system

nickoneill:
I think you can apply that `-skipPackagePluginValidation` in this PR, `xcargs` is a param that `scan` can take too, in the fastfile which is what the test runs in CI.

nickoneill:
OK, last change I think :) the project is trying to find the `GoogleService-Info.plist` file at `FiveCalls/FiveCalls/GoogleService-Info.plist` and the CI creates it at `FiveCalls/GoogleService-Info.plist`. If you adjust the path to look one directory higher then CI should be happy again

heathermh:
> OK, last change I think :) the project is trying to find the `GoogleService-Info.plist` file at `FiveCalls/FiveCalls/GoogleService-Info.plist` and the CI creates it at `FiveCalls/GoogleService-Info.plist`. If you adjust the path to look one directory higher then CI should be happy again

I think it was the reverse of this. I moved project's reference to plist one directory deeper than it was.

nickoneill:
Try applying this patch file? It moves the reference for the `GoogleService-Info.plist` up one level to where CI expects it. You'll probably have to move your local copy up one level too.

[0001-Move-google-info-plist-up-one-level.patch](https://github.com/5calls/ios/files/13260398/0001-Move-google-info-plist-up-one-level.patch)

I added you as a collaborator on this project as well, so if you create another branch on this project (not your fork), bitrise should run automatically.

PR #351, opened 2023-11-02
Done page design

nickoneill:
FYI on handling the call counts that I mentioned in a previous PR. I ended up removing the contacts from this page, I think it's superfluous info at this point.

There's also a crash here if you skip through the contacts too fast... still trying to figure out where that's coming from, the backtrace is not exactly clear.

Screenshot 2023-11-01 at 10 17 41 PM

Screenshot 2023-11-06 at 7 41 51 AM


FiveCalls/FiveCalls/IssueDone.swift
```diff
@@ -8,23 +8,161 @@
...
         }.navigationBarHidden(true)
         .clipped()
+        .frame(maxWidth: 500)
+        .onAppear() {
+            loadStats()
+        }
+    }
+        
+    func loadStats() {
+        let operation = FetchStatsOperation()
```
cselin, line 94: not required, but you could factor this out of the view using unidirectional flow allowing for better testability, maybe modifying the existing `FetchStats` to take an optional `issueID`?

cselin: Looks like that might actually remove the need for the view model entirely as you aren't currently using `showDonate`. Also simplifies having both an `issue` model and `viewModel`

cselin (approved):
Nice work, with a suggestion to integrate stats operation into existing unidirectional action

FiveCalls/FiveCalls/IssueDone.swift
nickoneill: Ugh, you're right. It's taking me a minute to make the unidirectional flow the default in my brain. Fixed

FiveCalls/FiveCalls/IssueDone.swift
```diff
@@ -46,25 +38,25 @@ struct IssueDone: View {
                     Spacer()
                 }.padding(.vertical, 16)
                 VStack {
-                    if let totalCalls = viewModel.totalCalls {
-                        CountingView(title: R.string.localizable.totalCalls(), count: totalCalls)
-                            .padding(.bottom, 14)
-                    }
-                    if let issueCalls = viewModel.issueCalls {
+                    CountingView(title: R.string.localizable.totalCalls(), count: store.state.globalCallCount)
```
cselin, line 41: Should we keep the `if let` in case this is zero?

PR #361, opened 2023-11-09
Issue list improvements

### LINKED ISSUE #362: issue order is not respected (see 5calls.org for proper ordering)

nickoneill:
Small change to how category view models are generated for the default list, this keeps the order of the json list.

Fixes #362 

~@cselin one other thing I was trying to accomplish here was scrolling to the top of the list when "more" or "less" is tapped. It doesn't seem like List has any way to set the scroll position, so I was considering moving to ScrollView and ForEach, then manually building a selection state for iPad. Happy to do that but I wanted to check if you can think of an easier way to accomplish the same thing~

### DIFF
--- a/FiveCalls/FiveCalls/Dashboard.swift
+++ b/FiveCalls/FiveCalls/Dashboard.swift
@@ -148,30 +148,37 @@ struct IssuesList: View {
     }
 
     var body: some View {
-        List(categorizedIssues, selection: $selectedIssue) { section in
-            Section {
-                ForEach(section.issues) { issue in
-                    NavigationLink(value: issue) {
-                        IssueListItem(issue: issue, contacts: store.state.contacts)
+        ScrollViewReader { scroll in
+            List(categorizedIssues, selection: $selectedIssue) { section in
+                Section {
+                    ForEach(section.issues) { issue in
+                        NavigationLink(value: issue) {
+                            IssueListItem(issue: issue, contacts: store.state.contacts)
+                        }
                     }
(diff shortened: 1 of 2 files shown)


cselin:
> Small change to how category view models are generated for the default list, this keeps the order of the json list.
> 
> @cselin one other thing I was trying to accomplish here was scrolling to the top of the list when "more" or "less" is tapped. It doesn't seem like List has any way to set the scroll position, so I was considering moving to ScrollView and ForEach, then manually building a selection state for iPad. Happy to do that but I wanted to check if you can think of an easier way to accomplish the same thing

It seems like [this should be supported](https://ishtiz.com/swiftui/how-to-scroll-to-a-specific-row-in-a-list-in-swiftui#:~:text=If%20you%20want%20to%20programmatically,scroll%20position%20of%20the%20list.), but I haven't been able to get it working yet. I can spend a little more time exploring since I think keeping list with section support, etc, would be nice/.

nickoneill:
Perfect, just pushed an update. Not sure if you were trying to scroll to a list index a la tableviewcontroller? Using the identifier of the issue worked for me.

cselin:
> Perfect, just pushed an update. Not sure if you were trying to scroll to a list index a la tableviewcontroller? Using the identifier of the issue worked for me.

I guess I misinterpreted it to use indexes, where ID's were necessary. I had this: `scrollView.scrollTo(0, anchor: .top)`

cselin (approved):

PR #364, opened 2023-11-11
Create an issue completion cache, show which calls have been made

### LINKED ISSUE #363: indicate which issues have been called about on the issue list (local issue completion cache is fine)

nickoneill:
Fixes #363 

Screenshot 2023-11-11 at 10 10 48 AM

### DIFF
--- a/FiveCalls/FiveCalls/AppState.swift
+++ b/FiveCalls/FiveCalls/AppState.swift
@@ -13,6 +13,16 @@ import os
 class AppState: ObservableObject, ReduxState {
     @Published var globalCallCount: Int = 0
     @Published var issueCallCounts: [Int: Int] = [:]
+    // issueCompletion is a local cache of completed calls: an array of contact ids keyed by an issue id
+    @Published var issueCompletion: [Int: [String]] = [:] {
+        didSet {
+            // NSNumber (bridged automatically from Int) is not supported as a key in a plist dictionary, so we stringify and unstringify
+            let plistSupportableIssueCache: [String: [String]] = Dictionary(uniqueKeysWithValues: issueCompletion.map { key, value in
+                (String(key), value)
+            })
+            UserDefaults.standard.set(plistSupportableIssueCache, forKey: UserDefaultsKey.issueCompletionCache.rawValue)
+        }
+    }
     @Published var donateOn = false
     @Published var issues: [Issue] = []
     @Published var contacts: [Contact] = []
@@ -43,16 +53,21 @@ class AppState: ObservableObject, ReduxState {
             
             switch locationType {
             case "address", "zipCode":
-                location = NewUserLocation(address: locationValue, display: locationDisplay)
+                self.location = NewUserLocation(address: locationValue, display: locationDisplay)
             case "coordinates":
                 let locValues = locationValue.split(separator: ",")
                 if locValues.count != 2 { return }
                 guard let lat = Double(locValues[0]), let lng = Double(locValues[1]) else { return }
                 
-                location = NewUserLocation(location: CLLocation(latitude: lat, longitude: lng), display: locationDisplay)
+                self.location = NewUserLocation(location: CLLocation(latitude: lat, longitude: lng), display: locationDisplay)
             default:
                 Logger().warning("unknown stored location type data: \(locationType)")
             }
+        }
+        
+        // load the issue completion cache
+        if let plistSupportableIssueCache = UserDefaults.standard.object(forKey: UserDefaultsKey.issueCompletionCache.rawValue) as? [String: [String]] {
+            self.issueCompletion = Dictionary(uniqueKeysWithValues: plistSupportableIssueCache.map({ key, value in (Int(key) ?? 0, value) }))
         }
     }
 }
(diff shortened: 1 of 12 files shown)


heathermh:
I think I'd like to know who I've called. What about possibly adding a checkmark overlay on each image, so I can see who it is and if I've contacted them?

This part would probably be a feature for later, but I know folks who are calling daily on some issues. It might be nice to have some indication of when the user called / how often.

FiveCalls/FiveCalls/AppState.swift
```diff
@@ -13,6 +13,16 @@ import os
 class AppState: ObservableObject, ReduxState {
     @Published var globalCallCount: Int = 0
     @Published var issueCallCounts: [Int: Int] = [:]
+    // issueCompletion is a local cache of completed calls: an array of contact ids keyed by an issue id
+    @Published var issueCompletion: [Int: [String]] = [:] {
+        didSet {
+            // NSNumber (bridged automatically from Int) is not supported as a key in a plist dictionary, so we stringify and unstringify
```
cselin, line 19: Does this also mean we couldn't leverage `@AppStorage` here?

FiveCalls/FiveCalls/AppState.swift
```diff
@@ -43,16 +53,21 @@ class AppState: ObservableObject, ReduxState {
...
-                location = NewUserLocation(location: CLLocation(latitude: lat, longitude: lng), display: locationDisplay)
+                self.location = NewUserLocation(location: CLLocation(latitude: lat, longitude: lng), display: locationDisplay)
             default:
                 Logger().warning("unknown stored location type data: \(locationType)")
             }
         }
+        
+        // load the issue completion cache
+        if let plistSupportableIssueCache = UserDefaults.standard.object(forKey: UserDefaultsKey.issueCompletionCache.rawValue) as? [String: [String]] {
+            self.issueCompletion = Dictionary(uniqueKeysWithValues: plistSupportableIssueCache.map({ key, value in (Int(key) ?? 0, value) }))
```
cselin, line 70: should we coalesce to 0 or just skip or error if we can't create a key?

cselin:
> I think I'd like to know who I've called. What about possibly adding a checkmark overlay on each image, so I can see who it is and if I've contacted them?
> 
> This part would probably be a feature for later, but I know folks who are calling daily on some issues. It might be nice to have some indication of when the user called / how often.

@heathermh you're referring to the issue detail, right? Yes, this was hooked up in the UIKit version.

Simulator Screenshot - iPad mini (6th generation) - 2023-11-11 at 18 54 44

cselin (approved):
Looks good with suggestion from @heathermh and potentially using AppStorage if possible, but just a nice-to-have really.

FiveCalls/FiveCalls/AppState.swift
nickoneill: Not that I can see, I can't get away from "no exact matches in call to initializer" if I use anything except a bare supported plist type (i.e. no arrays or dicts) for @AppStorage.

FiveCalls/FiveCalls/AppState.swift
nickoneill: Yeah, you're right. I'll drop anything that can't be converted to an Int.

nickoneill:
Updated with called indicators on the issuedetail page too:

Screenshot 2023-11-11 at 9 41 13 PM

heathermh:
> > I think I'd like to know who I've called. What about possibly adding a checkmark overlay on each image, so I can see who it is and if I've contacted them?
> > This part would probably be a feature for later, but I know folks who are calling daily on some issues. It might be nice to have some indication of when the user called / how often.
> 
> @heathermh you're referring to the issue detail, right? Yes, this was hooked up in the UIKit version.

@cselin In the summary screen where you have the stacked/overlapped photos - I think I'd like to see an indicator of which ones I've called right there. I like the version on the detail screen with image overlaid with checkmark. Could that fit on the summary screen with the tiny photos not overlapping as much? It would be nice to scan the list of issues and see that I've contacted someone for each that I care about.

cselin:
> > > I think I'd like to know who I've called. What about possibly adding a checkmark overlay on each image, so I can see who it is and if I've contacted them?
> > > This part would probably be a feature for later, but I know folks who are calling daily on some issues. It might be nice to have some indication of when the user called / how often.
> > 
> > 
> > @heathermh you're referring to the issue detail, right? Yes, this was hooked up in the UIKit version.
> 
> @cselin In the summary screen where you have the stacked/overlapped photos - I think I'd like to see an indicator of which ones I've called right there. I like the version on the detail screen with image overlaid with checkmark. Could that fit on the summary screen with the tiny photos not overlapping as much? It would be nice to scan the list of issues and see that I've contacted someone for each that I care about.

That's captured by Nick's screenshot in the description though, isn't it?

nickoneill:
Yeah, I think this is the right balance: glanceable idea of if you've completed all contacts on the dashboard, and if you need more info on rep name / bigger images, you can pop into the issue detail screen.

One thing I want to revisit: lots of people share the done screen on Instagram so I think we should add the contact summary from issue detail back into Done (it's there in UIKit, I originally had it in swiftui but removed it in favor of a simpler layout for this PR).

PR #372, opened 2023-11-18
Add plausible events back in

nickoneill:
Removed some old events we never ended up caring about, and removed tracking the contact pages (i.e. calling about issue x to contact y). We can construct that data from the call results if we want to, I'd rather stick to the basics with plausible, mostly how many unique people are flowing across all platforms.


cselin (approved):
🎉

PR #374, opened 2023-11-23
Reminders pass for accessibility

heathermh:
Made updates to make the reminders section more accessible. Focused on VoiceOver, dynamic text, lightness contrast for text, dark mode. Note that some SwiftUI bits don't play nice with dynamic text (especially the date picker and Done buttons).

One piece that needs @nickoneill's input in particular is the colors. I changed the colors fiveCallsDarkGray, fiveCallsMediumDarkGray, fivecallsRed. I changed them with use for text in mind - to get enough contrast when used as text over a white or black background. The grays might be fine, but the red ends up a little weird so I think you may want to adjust that one. Especially when we use it for such small text, it's hard to see against a white background and needed to be way lighter to work against black. IIRC the red is also used for a button somewhere, so we may want to have a red for decorative elements, and a different red (or different color) for text... maybe with a small red graphic instead?


nickoneill:
This looks great, thanks! I think the dark mode colors are something we can revisit when we are implementing dark mode so for now these placeholders with better contrast are fine.

If you rebase on top of the current main the CI errors should resolve themselves.

PR #375, opened 2023-11-28
Your Impact accessibility pass

heathermh:
* Is it OK if we use gear symbol for the menu? I think it's easier to see at this size than gear.circle.
* Updates for dynamic text, VoiceOver. Tried to keep general look of original. 
* It might make more sense to use a Back button in upper left/leading for My Impact and About (instead of Done button in upper right/trailing). I didn't make that change yet.

### DIFF
--- a/FiveCalls/FiveCalls/YourImpact.swift
+++ b/FiveCalls/FiveCalls/YourImpact.swift
@@ -38,7 +38,7 @@ struct YourImpact: View {
         }
     }
     
-    var showSubheading: Bool {
+    var showImpactList: Bool {
         userStats?.totalCalls() ?? 0 != 0
     }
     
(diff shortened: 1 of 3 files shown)


nickoneill (approved):
lgtm! Gear sans circle is good. The information hierarchy on this page isn't great, but that's not a problem with this PR, we can tackle in the future.

FiveCalls/FiveCalls/YourImpact.swift
```diff
@@ -49,54 +49,49 @@ struct YourImpact: View {
             
     var body: some View {
         NavigationStack {
-            VStack(spacing: 0) {
-                VStack(alignment: .leading, spacing: 15) {
-                    Text(weeklyStreekMessage)
+            List {
+                VStack(alignment: .leading, spacing: 16) {
+                    Text(weeklyStreakMessage)
```
cselin, line 54: Does including the streak message in the list have an side effects or is the layout the same?

FiveCalls/FiveCalls/YourImpact.swift
```diff
@@ -14,7 +14,7 @@ struct YourImpact: View {
 
     @State private var userStats: UserStats?
     
-    var weeklyStreekMessage: String {
+    var weeklyStreakMessage: String {
```
cselin, line 17: whoops, nice catch!

cselin (approved):

PR #376, opened 2023-11-29
About section accessibility pass

heathermh:
* Updated font colors to work for contrast and dark mode, use asset colors
* Each whole row of list is one big touch area, one VoiceOver section
* Updated text portions to resize with dynamic text and not get cut off


nickoneill (approved):
👍

Issue #377, opened 2023-11-30
Launch screen should match new layout


nickoneill:

nickoneill:
Fixed in #378

PR #378, opened 2023-11-30
Replace outdated launch screen storyboard with logo

nickoneill:
I would like to show the sidebar structure on iPad but I couldn't get varying the launch screen by platform working, I don't know if it's supported anymore as I can't find it in apple docs. This'll do to solve #377 for now and we can add a dark variation in the dark mode PR


heathermh (approved):
Cool for now

Issue #380, opened 2023-12-02
Ask for push permission on done page


nickoneill:

nickoneill:
Done in #381

PR #381, opened 2023-12-02
Re-integrated rating and push prompt on done

### LINKED ISSUE #379: Prompt for rating on the done page

nickoneill:
RatingPromptCounter is now static so we can use it without a view model on the done page. I would probably want some more concurrency checks if this was used in more places but it's just on the one page so should be fine.

Fixes #379, #380

### DIFF
--- a/FiveCalls/FiveCalls/IssueDone.swift
+++ b/FiveCalls/FiveCalls/IssueDone.swift
@@ -100,9 +100,10 @@ struct IssueDone: View {
         }.alert(R.string.localizable.notificationTitle(), isPresented: $showNotificationAlert) {
             Button {
                 // we don't really care which issue they were on when they subbed, just that it was a done page
-                AnalyticsManager.shared.trackEvent(name: "push-subscribe", path: "/issue/x/done/")
-                OneSignal.promptForPushNotifications(userResponse: { (success) in
-                    //
+                OneSignal.promptForPushNotifications(userResponse: { success in
+                    if success {
+                        AnalyticsManager.shared.trackEvent(name: "push-subscribe", path: "/issue/x/done/")
+                    }
                 })
             } label: {
                 Text(R.string.localizable.notificationImportant())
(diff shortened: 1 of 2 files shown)


FiveCalls/FiveCalls/IssueDone.swift
```diff
@@ -81,7 +85,58 @@ struct IssueDone: View {
...
+            // will prompt for a rating after hitting done 5 times
+            RatingPromptCounter.increment {
+                guard let currentScene = UIApplication.shared.connectedScenes.first as? UIWindowScene else {
+                    return
+                }
+                
+                SKStoreReviewController.requestReview(in: currentScene)
+            }
+            
+            // unlikely to occur at the same time as the rating prompt counter
```
cselin, line 98: in the worst case, what happens if they both trigger at the same time, stacked alerts?

FiveCalls/FiveCalls/IssueDone.swift
```diff
@@ -81,7 +85,58 @@ struct IssueDone: View {
...
+            checkForNotifications()
+        }.alert(R.string.localizable.notificationTitle(), isPresented: $showNotificationAlert) {
+            Button {
+                // we don't really care which issue they were on when they subbed, just that it was a done page
+                AnalyticsManager.shared.trackEvent(name: "push-subscribe", path: "/issue/x/done/")
+                OneSignal.promptForPushNotifications(userResponse: { (success) in
```
cselin, line 104:
```suggestion
                OneSignal.promptForPushNotifications(userResponse: { _ in })
```

cselin: unless you meant to handle the response

cselin (approved):

heathermh (approved):

FiveCalls/FiveCalls/IssueDone.swift
nickoneill: We actually only track this if the call succeeds in reminders, so I'll do the same here

FiveCalls/FiveCalls/IssueDone.swift
nickoneill: It's a bit jarring, one pops and then the other before you have a chance to react, but you can still interact with both in turn.

PR #383, opened 2023-12-02
Make welcome button tappable anywhere

### LINKED ISSUE #382: Welcome screen: button is only tappable on text, not blue background

nickoneill:
Move the button background inside the button so it's tappable anywhere, not just on the text.

Fixes #382


heathermh:
@nickoneill Thanks for getting back to this. Like a week ago I could have sworn there was a button that needed this and I couldn't remember where.

PR #392, opened 2023-12-13
Updated script replacements and tests

### LINKED ISSUE #391: Re-integrate script replacements

nickoneill:
Added script replacements for contact and location to contact detail page, and updated the tests since they're no longer on the Contact or Issue object.

Fixes #391

### DIFF
--- a/FiveCalls/FiveCalls/IssueContactDetail.swift
+++ b/FiveCalls/FiveCalls/IssueContactDetail.swift
@@ -23,6 +23,10 @@ struct IssueContactDetail: View {
         return Array(remainingContacts.dropFirst())
     }
     
+    var issueMarkdown: AttributedString {
+        return issue.markdownIssueScript(contact: currentContact, location: store.state.location)
+    }
+    
     var body: some View {
         ScrollView {
             VStack(alignment: .leading, spacing: 0) {
@@ -63,7 +67,7 @@ struct IssueContactDetail: View {
                         }
                     }
                 }.padding(.bottom)
-                Text(issue.markdownIssueScript(contact: currentContact, location: store.state.location))
+                Text(issueMarkdown)
                     .padding(.bottom)
                 if remainingContacts.count > 1 {
                     NavigationLink(value: IssueDetailNavModel(issue: issue, contacts: nextContacts)) {
--- a/FiveCalls/FiveCalls/IssueContactDetail.swift
+++ b/FiveCalls/FiveCalls/IssueContactDetail.swift
@@ -63,7 +63,7 @@ struct IssueContactDetail: View {
                         }
                     }
                 }.padding(.bottom)
-                Text(issue.markdownIssueScript)
+                Text(issue.markdownIssueScript(contact: currentContact, location: store.state.location))
                     .padding(.bottom)
                 if remainingContacts.count > 1 {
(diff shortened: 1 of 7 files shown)


FiveCalls/FiveCalls/IssueContactDetail.swift
```diff
@@ -63,7 +63,7 @@ struct IssueContactDetail: View {
                         }
                     }
                 }.padding(.bottom)
-                Text(issue.markdownIssueScript)
+                Text(issue.markdownIssueScript(contact: currentContact, location: store.state.location))
```
cselin, line 66: nit: can we migrate this to a computed property? Some benefits being better local testability (if we were ever to add unit tests here) and better adherence to observation if any of these other properties were to change. For example, if the location changed (however likely or unlikely this is), this text wouldn't be updated in its current form. In addition, here's an article explaining a good rule of thumb (use property for representing state or a value and use a method when invoking an action) when it comes to this: https://sarunw.com/posts/computed-property-vs-method/

FiveCalls/FiveCalls/ScriptReplacements.swift
```diff
@@ -0,0 +1,61 @@
...
+//  FiveCalls
+//
+//  Created by Nick O'Neill on 12/12/23.
+//  Copyright © 2023 5calls. All rights reserved.
+//
+
+import Foundation
+
+struct ScriptReplacements {
+    static func Replacing(script: String, contact: Contact, location: NewUserLocation?) -> String {
```
cselin, line 12:
```suggestion
    static func replacing(script: String, contact: Contact, location: NewUserLocation?) -> String {
```

cselin: nit: was capitalization intentional?

FiveCalls/FiveCalls/ScriptReplacements.swift
```diff
@@ -0,0 +1,61 @@
...
+//  ScriptReplacements.swift
+//  FiveCalls
+//
+//  Created by Nick O'Neill on 12/12/23.
+//  Copyright © 2023 5calls. All rights reserved.
+//
+
+import Foundation
+
+struct ScriptReplacements {
```
cselin, line 11: Touching this could be an opportunity to use the new Swift Regex system added in iOS 16. I might take this on as a follow up myself to get more familiar with it.

cselin (approved):
Looks good with a few suggestions

FiveCalls/FiveCalls/ScriptReplacements.swift
nickoneill: That would be great, I had no idea there was a more modern way to regex!

FiveCalls/FiveCalls/ScriptReplacements.swift
nickoneill: I was using the Go part of my brain to indicate which of these was exported, no reason to do that in swift though 🙃

FiveCalls/FiveCalls/IssueContactDetail.swift
nickoneill: my nemesis, thanks for catching this style issue again

Issue #393, opened 2023-12-14
Tipkit prevents Welcome screen from opening


nickoneill:

cselin:
I'll take a look at this today

PR #394, opened 2023-12-14
Handle deeplinks, move to swiftui-native main

### LINKED ISSUE #388: Support deeplinks
We already set `UserDefaultsKey.selectIssuePath` when the user arrives via a deeplink, we need to push the navigation stack on when we load issues if that key exists, perhaps in Dashboard?

nickoneill:
Handles deeplinks from the site, adopting the new `.onOpenURL` syntax in SwiftUI required a move to the SwiftUI-native Scene + WindowGroup creation in `App.swift`.

These changes introduce a bug between tipkit and the welcome screen: #393. There were already enough changes here that I didn't want to start messing with how tipkit is presented here.

Fixes #388

### DIFF
--- a/FiveCalls/FiveCalls/App.swift
+++ b/FiveCalls/FiveCalls/App.swift
@@ -14,18 +14,18 @@ struct FiveCallsApp: App {
             
     let store = Store(state: AppState(), middlewares: [appMiddleware()])
     @AppStorage(UserDefaultsKey.hasShownWelcomeScreen.rawValue) var hasShownWelcomeScreen = false
-    @State var welcomePresented = false
+    @State var showWelcomeScreen = false
 
     var body: some Scene {
         WindowGroup {
             IssueSplitView()
                 .environmentObject(store)
-                .sheet(isPresented: $welcomePresented) {
+                .sheet(isPresented: $showWelcomeScreen) {
                     Welcome().environmentObject(store)
                 }
                 .onAppear {
                     if !hasShownWelcomeScreen {
-                        welcomePresented = true
+                        showWelcomeScreen = true
                         hasShownWelcomeScreen = true
                     }
                 }
--- a/FiveCalls/FiveCalls/App.swift
+++ b/FiveCalls/FiveCalls/App.swift
@@ -8,14 +8,28 @@
 
 import SwiftUI
 
-//@main // TODO: once SwiftUI is done move main here
-//struct FiveCallsApp: App {
-//    @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
-//            
-//    var body: some Scene {
-//        WindowGroup {
-//            Dashboard().environmentObject(appDelegate.appState)
-//        }
-//    }
-//}
+@main
+struct FiveCallsApp: App {
+    @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
+            
+    let store = Store(state: AppState(), middlewares: [appMiddleware()])
+    @AppStorage(UserDefaultsKey.hasShownWelcomeScreen.rawValue) var hasShownWelcomeScreen = false
+    @State var welcomePresented = false
+
+    var body: some Scene {
+        WindowGroup {
+            IssueSplitView()
+                .environmentObject(store)
+                .sheet(isPresented: $welcomePresented) {
+                    Welcome().environmentObject(store)
+                }
+                .onAppear {
+                    if !hasShownWelcomeScreen {
+                        welcomePresented = true
+                        hasShownWelcomeScreen = true
+                    }
+                }
+        }
+    }
+}
 
--- a/FiveCalls/FiveCalls/AppDelegate.swift
+++ b/FiveCalls/FiveCalls/AppDelegate.swift
@@ -11,12 +11,7 @@ import SwiftUI
 import OneSignal
 import TipKit
 
-@UIApplicationMain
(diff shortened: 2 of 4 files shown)


FiveCalls/FiveCalls/App.swift
```diff
@@ -8,14 +8,28 @@
...
-//        }
-//    }
-//}
+@main
+struct FiveCallsApp: App {
+    @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
+            
+    let store = Store(state: AppState(), middlewares: [appMiddleware()])
+    @AppStorage(UserDefaultsKey.hasShownWelcomeScreen.rawValue) var hasShownWelcomeScreen = false
+    @State var welcomePresented = false
```
cselin, line 17: nit: we've been using a more active nomenclature for these types of properties, otherwise it sounded like a duplicate of `hasShownWelcomeScreen`
```suggestion
    @State var showWelcomeScreen = false
```

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -55,9 +38,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             let incomingURL = userActivity.webpageURL else {
             return false
         }
-        // sets a string like 'usps-postal-service-covid-funding' that we can use when issues are loaded
-        UserDefaults.standard.set(incomingURL.lastPathComponent, forKey: UserDefaultsKey.selectIssuePath.rawValue)
```
cselin, line 59: we don't need this anymore?

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -68,33 +49,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
     }
     
-    func transitionTo(rootViewController viewController: UIViewController) {
-        guard let window = self.window else { return }
-        guard window.rootViewController != viewController else { return }
-        
-        let snapshot = window.snapshotView(afterScreenUpdates: false)!
```
cselin, line 75: what was this snapshot stuff used for?

FiveCalls/FiveCalls/AppDelegate.swift
```diff
@@ -31,22 +26,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         oneSignalStartup(launchOptions: launchOptions)
         OneSignal.setExternalUserId(AnalyticsManager.shared.callerID)
         
-        window = UIWindow()
```
cselin, line 34: 🎉 eventually, we can probably migrate all of this out of the AppDelegate, right?

FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -60,6 +60,24 @@ struct Dashboard: View {
                 store.dispatch(action: .FetchContacts(location))
             }
         }
+        .onOpenURL(perform: { url in
+            if store.state.issues.isEmpty {
```
cselin, line 64: nit: it'd be great to move this logic to the reducer and middleware

FiveCalls/FiveCalls/AppState.swift
```diff
@@ -24,7 +24,11 @@ class AppState: ObservableObject, ReduxState {
         }
     }
     @Published var donateOn = false
-    @Published var issues: [Issue] = []
+    @Published var issues: [Issue] = [] {
+        didSet {
+            issueLoadedCallback?()
```
cselin, line 29: rather than require this temporary callback, couldn't we use middleware to do this only the first time issues are set?

cselin (approved):
Looks good with some suggestions

FiveCalls/FiveCalls/AppDelegate.swift
nickoneill: Nope, we get the url straight from `onOpenURL` and can do the same `lastPathComponent` parsing there

FiveCalls/FiveCalls/AppDelegate.swift
nickoneill: This was a custom UIKit transition between the welcome screen and the main screen. Not really noticeable, imo, I am interested in how to do the same thing on SwiftUI but not something needed in this location or for launch

FiveCalls/FiveCalls/AppDelegate.swift
nickoneill: Maybe in a few iOS versions? 😅 I don't think `App` supports 100% of what `UIApplicationDelegate` can handle yet

FiveCalls/FiveCalls/App.swift
nickoneill: 👍

FiveCalls/FiveCalls/Dashboard.swift
nickoneill: I attempted this but it didn't seem to fit with `selectedIssue` being a binding here. Would we move `selectedIssue` into the app state to accomplish this?

cselin: I think that would be reasonable, while we wouldn't be accessing the selected issue everywhere it is pretty instrumental to the flow of the app.

nickoneill: I can't figure out how to structure this right with `selectedIssue` currently on the router, I'm gonna merge and create an issue to take a look again in the future.

PR #397, opened 2023-12-15
Add phone number call and copy functionality

cselin:
Includes VoiceOver support (turn on sound to hear example):
https://github.com/5calls/ios/assets/810263/db495c95-120b-46f6-9c30-00f43fcf249c

Note:

- There's an issue where the field office name is considered a dimmed button that I haven't been able to fix
- It looks like we need to separate the call result buttons as right now it appears to be a single button


cselin:
@heathermh @nickoneill

nickoneill:
You missed some merge conflict text in `IssueContactDetail`

cselin:
> You missed some merge conflict text in `IssueContactDetail`

Thanks, fixed!

nickoneill (approved):
👍

heathermh:
@cselin Sorry for late reply on this. I was looking at it but got pulled away. If somebody gets back to this later, I think we should consider going back to the layout you had with phone number on the left and copy on the right. If we remove the section with the office name and adjust the number text to show office name (with or without the phone number), that gets rid of the dimmed button announcement from VoiceOver and keeps the dialing and copying functionality. We may need to update the accessibility labels a bit, but I think it could work well.

PR #402, opened 2023-12-19
Use swift regex for script customization

cselin:
Migrates script customization to use Swift Regex as well as cleaned up some more localized strings.


cselin:
@nickoneill reduced the regex code by quite a bit! Check out [this WWDC video](https://developer.apple.com/videos/play/wwdc2022/110357/) if you're interested to learn a bit more. They added builders similar to SwiftUI to make regex more declarative which we didn't have a use for here, but is pretty cool!

nickoneill (approved):
Awesome, much simpler!

Issue #398, opened 2023-12-17
Handle universal links in middleware rather than a view


nickoneill:
As part of #394, we get a universal link url, find any associated issue that's been loaded and assign that to `selectedIssue` in the `IssueRouter` to automatically navigate there.

@cselin suggested we do this via our actions/middleware and I agree, it's weird to need a callback when the issues are loaded when we could just handle this in the action that sets the issues when they're loaded. I took a stab at it during the original PR but I couldn't figure out a way to set `selectedIssue` in the router from the middleware/store since they're separate environment objects.

Do we need to refactor the router to be in the store now? Is there another way to deal with this that I'm missing?

cselin:
I think having the router be part of the store makes sense as that is supposed to be the full representation of state for the app.

PR #405, opened 2023-12-20
Migrate router to app state

cselin:
Migrates the router to the app state per this issue: https://github.com/5calls/ios/issues/398


FiveCalls/FiveCalls/Issue.swift
```diff
@@ -28,7 +28,7 @@ struct Issue: Identifiable, Decodable {
     }
     
     var shareURL: URL {
-        return URL(string: String(format: "https://5calls.org/issue/%s/",self.slug))!
+        return URL(string: String(format: "https://5calls.org/issue/%@/",self.slug))!
```
cselin, line 31: @nickoneill did the slugs recently change? I swear this worked with `%s` when I first tested, but then subsequently I was getting weirdly formatted share URL's soon after, but `%@` fixes it.

nickoneill: I moved this around in the last ~month, I probably formatted it like I was writing Go unintentionally

nickoneill (approved):
Awesome, thanks for tackling this

PR #406, opened 2023-12-22
Update phone number call/copy a11y

cselin:
Per @heathermh's suggestion [here](https://github.com/5calls/ios/pull/397#issuecomment-1867124852), I moved the office section to the call button and added office name to accessibility labels for clarity.
https://github.com/5calls/ios/assets/810263/928f3b06-c8f2-4fd9-9c0c-ce084dd82366


heathermh (approved):
I like how it turned out.

PR #407, opened 2023-12-30
Add a UI test for generating screenshots via fastlane

nickoneill:
Configures mock data for UI tests as well, the tricky part here is that URLSessionConfig is not something we can set during UI testing so unfortunately the mocking code has to live in the app itself and get configured by `launchEnvironment`. At least we can keep the mocked data json files in the test targets only.

UI tests are exclusively for screenshot generation for the moment, I find them too brittle for actual functional tests but open to adding some if other people want to generate some flows.

Tapping an outcome button doesn't appear to be accessible in the same way other buttons are, I suspect it's the same issue that @heathermh ran into when working on accessibility for those buttons (something to do with the lazy grid?) so this'll really be wrapped up once that fix is in.

Also removed the old screens from the repo, I don't think it's useful to check these in.

Updates #400


fastlane/Snapfile
```diff
@@ -1,17 +1,16 @@
 devices([
-  "iPhone 8",
-  "iPhone 8 Plus",
-  "iPhone 11",
-  "iPhone 11 Pro",
-  "iPhone 11 Pro Max",
-  "iPad Pro (12.9-inch) (3rd generation)"
+  # "iPhone SE (3rd generation)",
```
cselin, line 2: is this temporary? seems like we should still snapshot a variety of devices

fastlane/SnapshotHelper.swift
```diff
@@ -15,13 +15,12 @@
 import Foundation
 import XCTest
 
-var deviceLanguage = ""
-var locale = ""
-
+@MainActor
```
cselin, line 18: Does using `@MainActor` here slow down tests at all?

cselin (approved):
nice!

fastlane/Snapfile
nickoneill: Yes, I'll re-enable the relevant devices to get the needed screen sizes before merge

nickoneill:
I'm gonna take a look at the outcome buttons this evening and then update this PR to use those before merging this.

heathermh:
@nickoneill If the outcome buttons work well enough for screen shots, maybe this PR doesn't need to wait until they are fully functional?

fastlane/Snapfile
nickoneill: done

nickoneill:
Kinda regret this, I spent more time fussing with tests everywhere as I would have hand-rolling some screenshots. But here we are.

PR #411, opened 2024-01-03
Ensure welcome screen shown before menu popover tip

cselin:
Fixes the issue where menu popover tip prevents welcome screen from showing first time app is opened.

https://github.com/5calls/ios/assets/810263/c8426e49-70fb-4fae-920e-b06140eadca5

**Note:**
In this fix, we do not show the menu popover tip the first time the app is launched if the welcome screen is shown, but it will be shown subsequent app launches. I had another implementation that showed the tooltip after the welcome screen is shown, but it was buggy and often was delayed and sometimes even showed up on the detail screen when the menu isn't even available.

### DIFF
--- a/FiveCalls/FiveCalls/PopoverTip.swift
+++ b/FiveCalls/FiveCalls/PopoverTip.swift
@@ -21,18 +21,26 @@ struct PopoverTip: Tip {
 }
 
 extension View {
-    func popoverTipIfApplicable(title: Text, message: Text?) -> some View {
-    if #available(iOS 17, *) {
-        return self
-            .popoverTip(
-                PopoverTip(
-                    title: title,
-                    message: message
-                ),
-                arrowEdge: .top
-            )
-    } else {
-      return self
+    func popoverTipIfApplicable(showingWelcomeScreen: Bool,
+                                title: Text,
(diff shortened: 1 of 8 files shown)


FiveCalls/FiveCalls/PopoverTip.swift
```diff
@@ -21,18 +21,26 @@ struct PopoverTip: Tip {
...
-    } else {
-      return self
+    func popoverTipIfApplicable(showingWelcomeScreen: Bool,
+                                title: Text,
+                                message: Text?)
+        -> some View
+    {
+        if #available(iOS 17, *) {
+            if showingWelcomeScreen {
+                AnyView(self)
```
cselin, line 31: I spent a good hour trying to debug this method. If I kept the `return`'s, it compiled, but crashed without a helpful debug message which took me a while to track down. Adding `AnyView` seems to be required once I introduced the additional conditional 🤷🏻‍♂️

nickoneill: I ran into this before but didn't figure out how to get it working, `AnyView` to the rescue!

nickoneill (approved):

PR #414, opened 2024-01-04
Move the navigation link into the outcome grid

### LINKED ISSUE #413: Outcome buttons are a single VoiceOver element

nickoneill:
Looks like having a single nav link over a bunch of buttons was confusing VoiceOver. Making a nav link for each outcome button makes each button show in VoiceOver as expected.

Fixes #413

### DIFF
--- a/FiveCalls/FiveCalls/IssueContactDetail.swift
+++ b/FiveCalls/FiveCalls/IssueContactDetail.swift
@@ -142,22 +142,18 @@ struct IssueContactDetail: View {
                     .padding(.bottom)
                 
                 if remainingContacts.count > 1 {
-                    NavigationLink(value: IssueDetailNavModel(issue: issue, contacts: nextContacts)) {
-                        OutcomesView(outcomes: issue.outcomeModels, report: { outcome in
-                            let log = ContactLog(issueId: String(issue.id), contactId: currentContact.id, phone: "", outcome: outcome.status, date: Date(), reported: true)
-                            store.dispatch(action: .ReportOutcome(issue, log, outcome))
-                            store.dispatch(action: .GoToNext(issue, nextContacts))
-                        })
-                    }
+                    OutcomesView(outcomes: issue.outcomeModels, navModel: IssueDetailNavModel(issue: issue, contacts: nextContacts), report: { outcome in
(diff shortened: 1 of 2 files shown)


FiveCalls/FiveCalls/IssueContactDetail.swift
```diff
@@ -142,22 +142,18 @@ struct IssueContactDetail: View {
...
                 
                 if remainingContacts.count > 1 {
-                    NavigationLink(value: IssueDetailNavModel(issue: issue, contacts: nextContacts)) {
-                        OutcomesView(outcomes: issue.outcomeModels, report: { outcome in
-                            let log = ContactLog(issueId: String(issue.id), contactId: currentContact.id, phone: "", outcome: outcome.status, date: Date(), reported: true)
-                            store.dispatch(action: .ReportOutcome(issue, log, outcome))
-                            store.dispatch(action: .GoToNext(issue, nextContacts))
-                        })
-                    }
+                    OutcomesView(outcomes: issue.outcomeModels, navModel: IssueDetailNavModel(issue: issue, contacts: nextContacts), report: { outcome in
```
cselin, line 145: nit: can we clean up some of the duplication here? seems like the only difference between the two is whether to include `nextContacts` or not and the nav model.

cselin (approved):
LGTM with one suggestion

heathermh (changes requested):
4 buttons are seen by VO, but I can't tap them to advance to next screen.

I've been working on an option for this on another branch where the buttons do advance, but report(outcome) in OutcomesView doesn't work yet. (this code isn't pushed to GitHub yet)

I'd like to see OutcomesView take a more specific data type for navModel. I have a suggestion in process for this part, too.

heathermh:
Also just saw that when testing on device but not using VO for an issue with two contacts, the first contact runs OK but then it doesn't go to next contact... just goes straight to done page.

heathermh:
> Also just saw that when testing on device but not using VO for an issue with two contacts, the first contact runs OK but then it doesn't go to next contact... just goes straight to done page.
> 
> Update: when I test for an issue with 3 contacts, the first 2 work fine and the last one is skipped.

Update: this is a bug in the state... when I added it on my other branch, it does this. It's not a VO problem. I'm going to PR my branch and add an issue for this.

nickoneill:
closing to fix in #420

PR #415, opened 2024-01-04
Try to avoid slugs that don't make a valid URL

nickoneill:
I posted a version to TestFlight for the public over the weekend and found a crash for generating this share url from an invalid issue slug, we should do a better job of validating these on the server but let's be defensive against dumb mistakes on the client too

### DIFF
--- a/FiveCalls/FiveCalls/Issue.swift
+++ b/FiveCalls/FiveCalls/Issue.swift
@@ -28,7 +28,7 @@ struct Issue: Identifiable, Decodable {
     }
     
     var shareURL: URL {
-        return URL(string: String(format: "https://5calls.org/issue/%@/",self.slug))!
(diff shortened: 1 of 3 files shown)


FiveCalls/FiveCalls/Issue.swift
```diff
@@ -28,7 +28,7 @@ struct Issue: Identifiable, Decodable {
     }
     
     var shareURL: URL {
-        return URL(string: String(format: "https://5calls.org/issue/%@/",self.slug))!
+        return URL(string: String(format: "https://5calls.org/issue/%@/",self.slug.trimmingCharacters(in: .whitespacesAndNewlines)))!
```
cselin, line 31: Would be nice to have a unit test to ensure this issue doesn't crop up again in the future.

cselin (approved):
LGTM with one suggestion

FiveCalls/FiveCalls/Issue.swift
nickoneill: 👍 added a test

PR #419, opened 2024-01-06
Dashboard accessibility pass

heathermh:
Accessibility updates to allow Dashboard and subsequent views to work with VoiceOver, dynamic text

There was some phone call stuff in here early on, but I'm going to defer to cselin's updates already on main.


nickoneill:
There are a lot of changes here but I think a significant portion is stuff we already have squashed and merged so git is trying to apply it again. A rebase would fix but also be quite a process, I'm going to try to extract some individual changes by cherrypick + copypaste later today to make some smaller PRs out of this. I'll have some questions inline.

FiveCalls/FiveCalls/IssueDone.swift
```diff
@@ -140,25 +232,37 @@ struct CountingView: View {
...
+    }()
+
+    return IssueDone(issue: .basicPreviewIssue)
+        .environmentObject(Store(state: previewState, middlewares: [appMiddleware()]))
 }
 
-struct IssueNavModel {
-    let issue: Issue
+struct IssueDoneNavModel {
+    var issue: Issue
```
nickoneill, line 257: is there a reason to `var` this? I can't think of a reason to change the nav model after we've created and pushed it onto the nav stack

FiveCalls/FiveCalls/IssueListItem.swift
```diff
@@ -11,24 +11,28 @@ import SwiftUI
 struct IssueListItem: View {
     let issue: Issue
     let contacts: [Contact]
-    
+    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
+
+    func usingRegularFonts() -> Bool {
```
nickoneill, line 16: we could probably build a static typeutil for `usingRegularFonts` since it's being defined in multiple places

FiveCalls/FiveCalls/Localizable.strings
```diff
@@ -70,6 +70,8 @@
 
 // issue + issue contact detail screen
 "see-script"                          = "See your script";
+"main-phone"                          = "Washington, DC";
```
nickoneill, line 73: is this used somewhere? I don't see a reference to `main-phone`

FiveCalls/FiveCalls/LocationHeader.swift
```diff
@@ -44,14 +49,16 @@ struct LocationHeader: View {
...
-            .padding(.leading)
+            .padding(.horizontal, usingRegularFonts() ? 15 : 5)
             .padding(.vertical, 10)
-            Image(systemName: "location.circle")
-                .imageScale(.large)
-                .symbolRenderingMode(.hierarchical)
-                .font(.title3)
-                .padding(.trailing)
-                .padding(.leading, 7)
+            if usingRegularFonts() {
```
nickoneill, line 54: is the approach here to remove some iconography when dynamic type is set to a larger size? can you post some screenshots of how that changes the layout?

FiveCalls/FiveCalls/PreviewIssues.swift
```diff
@@ -27,4 +27,6 @@ extension Issue {
...
     static let multilinePreviewIssue = Issue(id: 812, meta: "", name: "Support the Act whose name is quite long", slug: "support-act-slug2", reason: Issue.issueReason, script: Issue.issueScript, categories: [Category(name: "Environment")], active: true, outcomeModels: [Outcome(label: "Contacted", status: "contact"), Outcome(label: "Voicemail", status: "voicemail")], contactType: "reps", contactAreas: ["US House", "US Senate"], createdAt: Date(timeIntervalSince1970: 1688015904))
+    static let extraLongPreviewIssue = Issue(id: 812, meta: "", name: "Call for supportive action in a made up region with extra long details", slug: "support-act-slug2", reason: Issue.issueReason, script: Issue.issueScript, categories: [Category(name: "Environment")], active: true, outcomeModels: [Outcome(label: "Contacted", status: "contact"), Outcome(label: "Voicemail", status: "voicemail")], contactType: "reps", contactAreas: ["US House", "US Senate"], createdAt: Date(timeIntervalSince1970: 1688015904))
```
nickoneill, line 30: the issue id should be distinct from any other preview issues, if you create a  list where items are identified by id and `812` shows up twice SwiftUI starts to behave real weird and it's hard to figure out what's wrong

FiveCalls/FiveCalls/IssueDone.swift
heathermh: Nope, it was a change that I didn't fully revert. Fixed now.

FiveCalls/FiveCalls/LocationHeader.swift
heathermh: In this case, the interface gets crowded with the largest fonts. Some of the text then gets truncated (which we could probably fix separately, if you want). Removing the location graphic helps to keep everything up higher / crowds out less of the issues list and ensures all the text is visible. Screen shots (12 pro, regular view w/ default size text, same with largest text, graphic removed for largest text):

Now I'm seeing that the 5 calls icon is still very small, which isn't ideal. But also, it doesn't add to the functionality.

regular

largeNoChange

largeImageRemoved

heathermh: You may also notice that for the largest fonts, the "What's important to you?" heading goes away. Screenshots (12 pro, very large text with header vs no header):
header

noHeader

FiveCalls/FiveCalls/PreviewIssues.swift
heathermh: I updated the ID to be unique. We could also consolidate to multilinePreviewIssue only, but use the longer name. I needed the longer name to be a better example of the data that's really out there.

FiveCalls/FiveCalls/LocationHeader.swift
nickoneill: 👍 this looks good, I usually see screens on social with a slightly larger text size, but not HUGE text like this so this seems like a decent tradeoff for the few people who need this giant size.

FiveCalls/FiveCalls/IssueListItem.swift
heathermh: Good in theory, but more complicated than I expected. I did try a few different ways, but getting runtime warnings about the environment variable.

PR #420, opened 2024-01-06
Heather's fix for outcome button accessibility

### LINKED ISSUE #413: Outcome buttons are a single VoiceOver element

nickoneill:
Implemented @heathermh's fix for outcome button accessibility, simplified the outcome nav per @cselin's request in #414, and fixed the off-by-one that was preventing the last contact from showing its IssueContactDetail.

Replaces #414, Fixes #413


nickoneill:
Updated the test to ensure correctness in the contact flow, agreed on the dispatch change but we can make it in a future PR when changes are closer to the middleware section

PR #421, opened 2024-01-08
Refactor NewUserLocation and LocationSheet to use async/await

cselin:
Includes the following:

- refactors usage of ~New~UserLocation to use async/await instead of callbacks
- disables detect location button and show activity indicator until location is found (or error) to prevent issues with continuation
- use existing error label to show other errors encountered with location search or detection
- added some tests

https://github.com/5calls/ios/assets/810263/7a1eed19-e652-46be-baa1-de4e077b2f57


cselin:
FYI @nickoneill here's a preview of converting `NewUserLocation` and `LocationSheet` to use async/await instead of closures (checkout first commit for now since 2nd is a large rename). I still have some debugging and testing to add, but so far it seems to simplify a lot.

FiveCalls/FiveCalls/LocationSheet.swift
```diff
@@ -108,9 +108,24 @@ struct LocationSheet: View {
     }
     
     func locationSearch() {
-        _ = NewUserLocation(address: locationText) { loc in
-            store.dispatch(action: .SetLocation(loc))
-            dismiss()
+        locationError = nil
+
+        Task {
```
cselin, line 113: These tasks can probably be refactored into an action and handled by middleware unless we want to keep them here because they are specific to this view only.

nickoneill: No strong feeling on it as long as it's only being used here

nickoneill (commented):
Looks great, thanks for tackling this! One of those transition changes that I had forgotten about

nickoneill (approved):
👍

PR #425, opened 2024-01-10
add code coverage to fastlane

cselin:
Adds code coverage lane which can be used to ensure minimum coverage
Screenshot 2024-01-11 at 9 58 21 AM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,4 +6,4 @@ gem "cocoapods-acknowledgements"
 gem "dotenv"
 gem "fastlane"
 gem "badge"
-
+gem "xcov"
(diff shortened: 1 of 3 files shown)


cselin:
@nickoneill this was easier than I thought using the (old) xcov tool with fastlane. I've verified this works locally and errors if minimum coverage goes below 24.0%. The only missing piece is how to integrate this with Bitrise which I don't seem to be able to tap into for this app even on a fork, it goes to your workspace.

cselin:
Seems like we can probably remove `scan` from `coverage` lane and just update `fastlane "ios" "test"` -> `fastlane "ios" "test" "coverage"`

nickoneill:
changed the bitrise invocation, let's see if it does the right thing for this branch

nickoneill (approved):
needs two steps, `ios test` and `ios coverage` but it works great

PR #427, opened 2024-01-20
Tweak contact's call/copy to adjust for AX text sizes

heathermh:
Menu formatting doesn't want to play nice. Added check for AX1 and larger text so we can make sure user knows they're calling vs copying the phone number

Checked animations and I think it's OK to leave them for now. They're minimal and hopefully not intrusive for folks who have reduce motion enabled.

### DIFF
--- a/FiveCalls/FiveCalls/IssueContactDetail.swift
+++ b/FiveCalls/FiveCalls/IssueContactDetail.swift
@@ -10,6 +10,7 @@ import SwiftUI
 
 struct IssueContactDetail: View {
     @EnvironmentObject var store: Store
+    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
 
     let issue: Issue
     let remainingContacts: [Contact]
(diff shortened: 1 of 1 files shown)


cselin:
Sounds good, but can you include a screenshot of how this looks?

heathermh:
> Sounds good, but can you include a screenshot of how this looks?

Here's how it will look default sizes vs AX1 vs AX5 By default, Menu will remove the symbols whenever we use AX sizes so I adjusted the text so it still makes sense. I don't see a way to fix the text cutting off, though. It's not perfect, but I think it's functional for most use cases. Users can also copy the phone number and paste elsewhere if they want to see full phone number. 

callDefault

callAX1

callAX5

heathermh:
@cselin I'm testing this in the simulator for iPad and seeing inconsistent behavior (no copy column in the menu at all). Do you have a hardware device to check this on? If not, I think I can scrape up an old iPad; will have to look.

heathermh:
Update: I found an iPad to test with and this menu isn't working on iPad. Added new issue https://github.com/5calls/ios/issues/429

cselin:
Thanks for providing screenshots, I understand the issue now. While it wouldn't minimize the cutoff completely for longer city names, we could drop the phone number from all these columns since adding "Call" or "Copy" should make it obvious what it does. I can take a look at the iPad issue once you get this in.

cselin (approved):

cselin:
Actually sorry, my iPad is too old to run 16.2 :-/

nickoneill:
:+1: looks good, feel free to merge

heathermh:
Merging this in. Open to making more changes, but I think this issue is resolved enough for 2.0.

PR #431, opened 2024-01-25
Add delay for iOS 17+ to avoid race between popover tip and location sheet

cselin:
https://github.com/5calls/ios/assets/810263/4f1097ac-fee5-49a2-b519-bba05aff41ac

### DIFF
--- a/FiveCalls/FiveCalls/Dashboard.swift
+++ b/FiveCalls/FiveCalls/Dashboard.swift
@@ -25,7 +25,13 @@ struct Dashboard: View {
                 LocationHeader(location: store.state.location, fetchingContacts: store.state.fetchingContacts)
                     .padding(.bottom, 10)
                     .onTapGesture {
-                        showLocationSheet.toggle()
+                        if #available(iOS 17.0, *) {
+                            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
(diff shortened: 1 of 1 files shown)


FiveCalls/FiveCalls/Dashboard.swift
```diff
@@ -25,7 +25,13 @@ struct Dashboard: View {
                 LocationHeader(location: store.state.location, fetchingContacts: store.state.fetchingContacts)
                     .padding(.bottom, 10)
                     .onTapGesture {
-                        showLocationSheet.toggle()
+                        if #available(iOS 17.0, *) {
```
cselin, line 28: I started going down a route to try and track whether the popover tip is showing (or at the very least eligible to be showing), but it got tricky due to it being restricted to iOS 17+ so I couldn't just store the tip in the store state.

cselin: There isn't a a hugely noticeable delay when tapping the button, so hopefully this is reasonable.

nickoneill (approved):
Starting to regret tipkit v1 yet? 😂

cselin:
> Starting to regret tipkit v1 yet? 😂

Yeaaaa, maybe it's a little too bleeding edge

PR #441, opened 2024-02-10
Remove button icons and padding so outcomes don't wrap

### LINKED ISSUE #436: Reduce text size before breaking lines on outcome buttons
Between small size screens and various font modifiers, I've seen a number of screenshots where the outcome buttons break weirdly. We should improve their behavior so they reduce their font size to fit (the font size should reduce on all buttons, not just breaking ones) if we detect they may be too long.

Screenshot 2024-02-07 at 8 24 18 AM

nickoneill:
I tried using `.minimumScaleFactor` here to reduce text size when it's too large which was great but I couldn't figure out how to match that size across all buttons, which is the design I would want to see for it.

Instead, just remove some icons and padding so it fits on small screens.

Fixes #436 

Previously:
303403604-d064f3e5-5fe1-4cc4-9d4d-0841ffa02b27

fixed:
Screenshot 2024-02-11 at 8 28 55 AM

### DIFF
--- a/FiveCalls/FiveCalls/OutcomesView.swift
+++ b/FiveCalls/FiveCalls/OutcomesView.swift
@@ -15,8 +15,7 @@ struct OutcomesView: View {
     var body: some View {
         LazyVGrid(columns: [GridItem(.flexible()),GridItem(.flexible())]) {
             ForEach(outcomes) { outcome in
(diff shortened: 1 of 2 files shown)


cselin (approved):
LGTM. Mind adding before and after screenshots to document changes?

nickoneill:
added!

PR #446, opened 2024-02-24
Remove ControlGroup < iOS 16.4 to fix area office numbers

cselin:
While debugging I noticed that if `ControlGroup` was used at all within `Menu` it wouldn't render anything inside of it so I refactored to remove it for iOS < 16.4 since we know its broken on 16.2 and I don't have 16.3 to test. I ensured VoiceOver still works including setting focus to copied text (at least on my device iOS 17.2.1 but it should work as well as before with the new UI <16.4).

|>= iOS 16.4 (no change)|< iOS 16.4|
|-|-|
|IMG_2782|Simulator Screenshot - iPhone 12 - 2024-02-25 at 07 25 05|

### DIFF
--- a/FiveCalls/FiveCalls/IssueContactDetail.swift
+++ b/FiveCalls/FiveCalls/IssueContactDetail.swift
@@ -10,7 +10,6 @@ import SwiftUI
 
 struct IssueContactDetail: View {
     @EnvironmentObject var store: Store
-    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
 
     let issue: Issue
     let remainingContacts: [Contact]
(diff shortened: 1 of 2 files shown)


nickoneill:
Thanks! I have been fielding a few support issues about this as well. Duplicating copy and call lines for each office feels real busy, I wonder if we should limit the options to just "call" for the iOS 16.2 menu? I'm glad we have copy for the control group menu but call is without a doubt the most used option vs copy and pasting the number elsewhere.

cselin:
> Thanks! I have been fielding a few support issues about this as well. Duplicating copy and call lines for each office feels real busy, I wonder if we should limit the options to just "call" for the iOS 16.2 menu? I'm glad we have copy for the control group menu but call is without a doubt the most used option vs copy and pasting the number elsewhere.

Good call. I removed copy <16.4 and restored the icon for <16.2 call instead of using "Call <Place> <Number>"

nickoneill (approved):
👍 thanks for diving into this tricky spot!

PR #448, opened 2024-02-25
Refresh issues when they haven't been fetched in an hour

### LINKED ISSUE #444: Refresh issues list if reopened after x hours
In the switch to SwiftUI we lost the automatic pull-to-refresh control on the issues list - I don't actually miss this, it's a pretty hidden interaction and we don't push new issues regularly enough where people _should_ be refreshing the list manually. Our app is also pretty lean and it can live in memory on iOS for a long time, leading to opening the app days later and not seeing issue updates. 

We should keep track of the last time the issues list was loaded and refresh it if the app is opened from a suspended state after ~6 hours.

### LINKED ISSUE #447: Issues fail to load rarely
@cselin saw this once during development and I got a support issue about it the other day, otherwise not reproducible. The user loaded the app and no issues appeared in the list. I asked them to force quit the app and relaunch, they indicated that this **did not** fix the issue but a reinstall from the App Store resolved the issue.

Questionable if the user did a force quit correctly, my brain is coming up with nothing that would work after a reinstall but not after a force quit.

nickoneill:
Moves issue refresh to scene phase changes and out of the dashboard component (@cselin will like this), and tests when the app comes back to foreground.

Fixes #444, maybe will resolve #447 if the cause is something weird about `onAppear` not firing?

### DIFF
--- a/FiveCalls/FiveCalls/AppState.swift
+++ b/FiveCalls/FiveCalls/AppState.swift
@@ -26,6 +26,7 @@ class AppState: ObservableObject, ReduxState {
     }
     @Published var donateOn = false
     @Published var issues: [Issue] = []
+    @Published var issueFetchTime: Date? = nil
     @Published var contacts: [Contact] = []
     @Published var location: UserLocation? {
         didSet {
@@ -90,5 +91,17 @@ extension AppState {
         }
 
         return contactIDs.contains(contactID)
+    }
+    
+    var needsIssueRefresh: Bool {
+        guard let issueFetchTime else {
+            return true
+        }
+        
+        if issueFetchTime < Date().addingTimeInterval(-1 * 60) {
+            return true
+        } else {
+            return false
+        }
     }
 }
--- a/FiveCalls/FiveCalls/Dashboard.swift
+++ b/FiveCalls/FiveCalls/Dashboard.swift
@@ -67,11 +67,6 @@ struct Dashboard: View {
         .onAppear() {
             AnalyticsManager.shared.trackPageview(path: "/")
 
-            // TODO: refresh if issues are old too?
-            if store.state.issues.isEmpty {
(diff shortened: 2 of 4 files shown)


FiveCalls/FiveCalls/AppState.swift
```diff
@@ -91,4 +92,13 @@ extension AppState {
 
         return contactIDs.contains(contactID)
     }
+    
+    var needsIssueRefresh: Bool {
+        if issueFetchTime == nil {
```
cselin, line 97: A little less prone to error if we ever make changes to do something like:
```
guard let issueFetchTime else {
   return true
}

if issueFetchTime < Date().addingTimeInterval(-1 * 60) {
    return true
} else {
    return false
}

File: FiveCalls/FiveCalls/Dashboard.swift [hunk 2]
```diff
@@ -67,11 +67,6 @@ struct Dashboard: View {
         .onAppear() {
             AnalyticsManager.shared.trackPageview(path: "/")
 
-            // TODO: refresh if issues are old too?
-            if store.state.issues.isEmpty {
```
cselin, line 71: So the `active` scene phase change is always triggered upon launch?

cselin (approved):
Looks good with some suggestions

FiveCalls/FiveCalls/Dashboard.swift
nickoneill: Yep, active happens when an app becomes frontmost and interactive, so the common cases would be startup and returning from the background. It can happen pretty rapidly when multitasking but I think that's OK given the timer on fetching issues again

FiveCalls/FiveCalls/AppState.swift
nickoneill: Good call, much easier to think about this way

PR #450, opened 2024-08-30
Update LocationSheet.swift

yinuoyang01:
Fix accessibility issues of the magnifying glass search button

### DIFF
--- a/FiveCalls/FiveCalls/Localizable.strings
+++ b/FiveCalls/FiveCalls/Localizable.strings
@@ -110,7 +110,7 @@
 "enter-location"                      = "Enter a location";
 "unknown-location"                    = "Unknown Location";
 "detect-location"                     = "Detect my location";
-"search-location"                     = "Search my location"
+"search-location"                     = "Search my location";
 "location-instructions"               = "Use an address, zip code or zip + 4";
 "location-error-off"                  = "Location permission is off";
 "location-error-default"              = "An error occured trying to find your location";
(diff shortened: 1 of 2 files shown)


nickoneill:
Thank you! Can you add these strings into the localization files and use the generated `R.localization` strings in the code?

yinuoyang01:
> Thank you! Can you add these strings into the localization files and use the generated `R.localization` strings in the code?

Hi, do you mean add the new strings to the corresponding language files to support more languages?

yinuoyang01:
Hi, I have change the text to use the generated R.localization strings.

nickoneill:
We don't need translations for other languages, but we store all our strings in `Localizable.strings` so that if we want to localize in the future, we can simply translate that file to start the process.

Currently you set the accessibility label to `R.string.localizable.detectLocation()` which is `detect-location` in `Localizable.strings`, so it would read "Detect my location" which is incorrect for that button. I think your original label is correct "Search Location", you just need to add a new line in `Localizable.strings` for this (match the styles already in that file for the name) and then replace the usage with `R.string.localizable.whateverYouCalledYourString()`

yinuoyang01:
Got it, I have added the string to Localizable.strings and replace the usage.

nickoneill (approved):

nickoneill:
Thanks! We'll get this into the next release

nickoneill:
Sorry, one more thing: you need a semicolon at the end of the line in the strings file @yinuoyang01

yinuoyang01:
> Sorry, one more thing: you need a semicolon at the end of the line in the strings file @yinuoyang01

Hi, sorry to miss that, I have added the semicolon!

nickoneill:
Thanks!

Issue #451, opened 2024-11-11
Swipe to go back is not working


nickoneill:
I'm not sure what happened to this, I think we lost it in the SwiftUI transition. It feels like we're doing navigation in more or less the standard way, so I don't know what is happening here.

We should be able to swipe back from issue detail -> issue list at the very least, but ideally any of the back buttons in the call flow should also be swipe-backable (but not when you're at the done screen where there is no back button).

nickoneill:
fixed in #456

PR #456, opened 2024-11-25
Issue #451: Swipe to go back is not working

iosjillian:
* Fixes Issue #451 
* Removed custom back navigation
* Set up navigation destinations
* Updated README
* Removed cocoapods from Gemfile

Here's what it looks like! Not sure why there's a circle on the Chevron, maybe that's a new default? I'm happy to make updates to the UI if desired. This also changed the opacity/scroll behavior of the nav bar a bit to the default.

https://github.com/user-attachments/assets/3555fff1-49d5-46a7-a06c-b5658c7b4fc3


cselin (approved):
Nice work!

FiveCalls/FiveCalls/IssueDone.swift
```diff
@@ -115,7 +115,6 @@ struct IssueDone: View {
             }
             .padding(.horizontal)
         }
-        .navigationBarHidden(true)
```
nickoneill, line 118: Let's keep this line, I don't think you should be able to go back from the done page

nickoneill (approved):
Good stuff! The custom header was not doing enough to be worth the extra complexity

FiveCalls/FiveCalls/IssueDone.swift
iosjillian: Good catch! Updated

nickoneill (approved):
excellent, thank you!

PR #460, opened 2024-12-01
Fixes checkmarks for contactids with dashes

### LINKED ISSUE #457: Contacts are not checked as completed for certain types
A user reports that Governors and State Rep contacts don't change their icons to checkmarks when you've completed calls in the app. House and Senate seem to work fine though.

nickoneill:
Fixes #457

We were assuming all contactIDs had no dashes. We were wrong. Splits before the last appearing dash to get the contactid now, added some tests.

### DIFF
--- a/FiveCalls/FiveCallsTests/AppStateTests.swift
+++ b/FiveCalls/FiveCallsTests/AppStateTests.swift
@@ -1,0 +1,76 @@
+import XCTest
+@testable import FiveCalls
+
+final class AppStateTests: XCTestCase {
+    
+    var sut: AppState!
+    
+    override func setUp() {
+        super.setUp()
+        sut = AppState()
+    }
+    
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
+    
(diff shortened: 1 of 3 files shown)


FiveCalls/FiveCallsTests/AppStateTests.swift
```diff
@@ -0,0 +1,76 @@
+import XCTest
+@testable import FiveCalls
+
+final class AppStateTests: XCTestCase {
+    
+    var sut: AppState!
```
cselin, line 6: _nit:_ what is "sut"?

cselin (approved):
Great tests!

FiveCalls/FiveCallsTests/AppStateTests.swift
nickoneill: **s**tate **u**sed for **t**est? 😂 I have no clue, I asked Claude to generate a test harness for the app state.

cselin: Haha, well that's exciting!

PR #469, opened 2025-03-22
Remove local html content

nickoneill:
Keeping the local HTML content up-to-date separately was a pain and unnecessary since we have it on the site. For about us and why calling works, just open the relevant website links in safari.

(also maybe removes the last displayed twitter link in the project?)


dektar (approved):
LG although I know absolutely nothing about ios/swift!

PR #470, opened 2025-03-22
Indicate when a user has entered a split zip

nickoneill:
Using our new `isSplit` in the contacts response, we tell the user if they've entered a split district. They can still use it if they want!

https://github.com/user-attachments/assets/49e4006c-3ab6-414f-98ec-86d25a428afb

### DIFF
--- a/FiveCalls/FiveCalls/LocationHeader.swift
+++ b/FiveCalls/FiveCalls/LocationHeader.swift
@@ -10,6 +10,7 @@ import SwiftUI
 
 struct LocationHeader: View {
     let location: UserLocation?
+    let isSplit: Bool
     let fetchingContacts: Bool
     @Environment(\.dynamicTypeSize) private var dynamicTypeSize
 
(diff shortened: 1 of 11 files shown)


FiveCalls/FiveCalls/LocationHeader.swift
```diff
@@ -90,9 +99,17 @@ struct LocationHeader: View {
 struct LocationHeader_Previews: PreviewProvider {
     static var previews: some View {
         VStack {
-            LocationHeader(location: nil, fetchingContacts: true)
-            LocationHeader(location: nil, fetchingContacts: false)
-            LocationHeader(location: UserLocation(address: "19444"), fetchingContacts: false)
+            LocationHeader(location: nil, isSplit: false, fetchingContacts: true)
+            LocationHeader(location: nil, isSplit: true, fetchingContacts: false)
+            LocationHeader(
+                location: UserLocation(address: "19444"),
```
dektar, line 105: i don't know what this preview thing is about -- is this just for development? seems weird to hard-code here as zip codes can be redistricted.

dektar (commented):
LG % failing check. Also if this were work I'd ask for test coverage but have no leg to stand on in 5calls land as I barely have any in android!

dektar (approved):

FiveCalls/FiveCalls/LocationHeader.swift
nickoneill: Yep, these are a set of view configurations that will show in Xcode so you can see how the different settings are rendered, development only. They don't even depend on the zip in question being split because at this point down the view hierarchy we're just describing how to render so we declare "this should be shown as split" or not

Issue #466, opened 2025-03-14
Change photo circles colors to represent party


nickoneill:
If we can't get the photos to load (#465 or we don't have a photo yet), default to circles with party-standard colors: a red shade for Republicans, a blue shade for Democrats and a purple shade for Independents.

raysj9:
Just checking to see if you are still looking for someone to take this on. If so, I'm happy to help!

nickoneill:
Yep, @raysj9, please jump in if you are interested! Just make a mention here that you'll take a look at this so someone else doesn't work on it at the same time.

raysj9:
Great. I'll take this

nickoneill:
fixed in #471

Issue #465, opened 2025-03-14
AsyncImage photos frequently don't load


nickoneill:
We're just using bog standard AsyncImage views on the main topic list and issue details and very frequently the photos will fail to load and just show the default colored circles. They'll load fine elsewhere in the same session so I don't know what's going on. These are definitely reps with available photos.

raysj9:
I'll take this

PR #471, opened 2025-04-04
Issue #466: Change photo circles colors to represent party

raysj9:
- Addresses https://github.com/5calls/ios/issues/466
- When a photo does not load or is unavailable, the respective party color is displayed instead of the sample colors.
- If no party information is available, the fallback remains the sample colors.
- Used soft tones of red, blue, and purple to match the aesthetic of the sample colors, but this can be changed if needed.

https://github.com/user-attachments/assets/d64ed8a6-2b8e-4353-8408-0268a8dffe90

Example of person with no party affiliation.


nickoneill (approved):
Thanks! Any chance you want to take a look at #465 too? This is a good fallback but honestly it shouldn't happen as frequently as it does

raysj9:
@nickoneill Agreed. I'll take a look

PR #472, opened 2025-04-08
Issue #465: fix async image loading issue

raysj9:
Fixes https://github.com/5calls/ios/issues/465

It seems this bug appears only when the app is in a compact size class (on an iPhone or a smaller iPad window). On a full-screen iPad, when both columns of the split view are visible, it does not occur.

Based on additional testing in a standalone playground, it seems this is just a bug with AsyncImage in a split view context. (https://gist.github.com/raysj9/5f1b89fff62b5d8145323eec21937d21)

This workaround forces the issue detail view to refresh, which allows the Async images to load reliably. I'm thinking the only other way around this would be to find another way to load images without using a standard Swiftui async image.

Before: 

https://github.com/user-attachments/assets/f56c4317-52f1-46d6-bf97-bec329776b49

After:

https://github.com/user-attachments/assets/47152b00-ed95-430d-800a-666431d91d64

### DIFF
--- a/FiveCalls/FiveCalls/IssueDetail.swift
+++ b/FiveCalls/FiveCalls/IssueDetail.swift
@@ -70,10 +70,6 @@ struct IssueDetail: View {
             .padding(.horizontal)
         }
         .id(forceRefreshID)
-        .onAppear {
-            // Force refresh id so that async images can load. See https://github.com/5calls/ios/issues/465
-            forceRefreshID = UUID()
-        }
         .navigationBarTitleDisplayMode(.inline)
         .toolbar {
             ToolbarItem(placement: .navigationBarTrailing) {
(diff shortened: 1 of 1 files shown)


nickoneill:
Wow, that's not what I expected at all but great catch. There are definitely some alternative async image views we could use but I'm not sure it's worth it if this little tweak fixes it.

The cost of the fix is that whenever the view transitions to be the frontmost view, it rerenders?

raysj9:
Yes. Whenever this view appears, it will re-render, but the view is simple enough that performance shouldn’t be an issue. 

One small side effect is that the scroll position resets when navigating away and coming back, though most users likely won’t mind. Not ideal, but better than images failing half the time.

PR #475, opened 2025-05-23
Only rerender contact item

raysj9:
Still just a workaround for #465 #472 but only rerenders the contact list item rather than the whole view itself. Should fix the issue where the scroll position would reset every time you navigated back to the issue detail.


raysj9:
We did, but I noticed we were unnecessarily rerendering the whole screen rather than just the part with the images. It's just a minor tweak.

nickoneill (approved):
👍

Issue #473, opened 2025-04-08
Handle vacant seats better


nickoneill:
Right now when a seat is vacant (a member has died and no special election for their seat has occurred yet), we display the "set your location" message because there's no contact returned for that area.

We should do a better job of messaging that there isn't a rep for this area and nothing is wrong with your location.

raysj9:
Is this issue accepting help? Probably worth looking at given the number of vacancies currently.

raysj9:
Cool. I'll take a look at it

nickoneill:
fixed in #477

PR #477, opened 2025-05-25
Handle vacant senate, house reps

raysj9:
This handles #473. It assumes that the seats are vacant whenever there is no house rep or less than 2 senators. If this is not always the case, the wording can be changed.

@nickoneill Do you think that missing Governors and Attorneys General should be handled as well? Not sure how common that is.

### Issue for Senators, House Reps, but missing House Rep

<p>
  Simulator Screenshot
</p>

### Issue for House Reps only, but missing House Rep

Prior behavior was to show the Set Your Location button

<p>
  Simulator Screenshot
</p>

### Location Not Set

<p>
  Simulator Screenshot
</p>

### DIFF
--- a/FiveCalls/FiveCalls/IssueDone.swift
+++ b/FiveCalls/FiveCalls/IssueDone.swift
@@ -231,6 +231,10 @@ struct CountingView: View {
             return 2000000
         } else if count < 4500000 {
             return 5000000
+        } else if count < 9500000 {
+            return 10000000
+        } else if count < 12500000 {
+            return 13000000
         }
 
         return 0


FiveCalls/FiveCalls/IssueDetail.swift
```diff
@@ -40,19 +44,26 @@ struct IssueDetail: View {
                             NavigationLink(value: IssueDetailNavModel(issue: issue, contacts: Array(contacts[contact.number..<contacts.endIndex]))) {
                                 ContactListItem(contact: contact.element, showComplete: store.state.issueCalledOn(issueID: issue.id, contactID: contact.id))
                             }
-                            if contact.number < contacts.count - 1 {
+                            if contact.number < contacts.count - 1 || contact.number == contacts.count - 1 && !vacantAreas.isEmpty {
```
nickoneill, line 47: I realize this is a stretch but this only handles the case of a single vacancy, right? We don't have to handle the case for multiple vacancies since it's so rare, but I think it would help to better describe what this is doing ("either are not done with the contact list or it's the last contact and we have a vacancy to show")

nickoneill (changes requested):
Thank you! One question in-line about clarity.

Can we show the same thing for the Reps tab?

nickoneill (approved):

Issue #476, opened 2025-05-24
Show House and Senate contacts always, indicate relevancy


nickoneill:
Per https://github.com/5calls/5calls/issues/78, we should show both House and Senate reps when the issue targets either house or senate. If the call does not direct that we should call the House, we should dim the contact and adjust the subtitle to indicate that this rep is not relevant for this call at this time.

See the parent issue for a mock for the web.

raysj9:
I'll take this one

PR #478, opened 2025-06-12
Always show house senate reps

raysj9:
- Implements #476.
- Built on top of the work in #477, which was approved but not yet merged. Since this PR includes all of those changes  we may not need to merge that PR separately.
- Contacts for both house and senate will now be shown on issues, regardless of relevancy.
- [Refactored `IssueDetail`](https://github.com/5calls/ios/compare/main...raysj9:ios:always-show-house-senate-reps?expand=1#diff-7e192fca51c9fed27282efcf53830fe83534861f538bb7e9ff0a016fcc9c3d46R48-R62) layout to avoid a SwiftUI compiler error ("The compiler is unable to type-check this expression in reasonable time...").

Simulator Screenshot


nickoneill:
Thanks @raysj9, sorry for the wait on this review but this looks great. Can you rebase to fix the conflicts and we can merge? It should be straightforward, we squash changes so the vacant seat changes are there but not the exact same commits that exist in this branch.

PR #481, opened 2025-07-04
Remove dead code

Abizern:
Xcode flagged an unused variable, and raised a warning. It may have been left behind following a refactoring.


nickoneill:
Thanks, good call. I usually think of warnings as errors in Xcode, I wonder if we could formalize that for the tests we run on PRs?

Issue #484, opened 2025-07-12
Use script customization endpoint


nickoneill:
Use the new script API endpoint to display custom scripts for each contact. See parent for details on the endpoint.

PR #485, opened 2025-08-20
Script customization

raysj9:
- Implements #484
- For issues that support it, scripts will be customized server-side, rather than client-side, based on the user's representatives and location. 
- The added fetch operation supports an optional caller name, but is unused since the app does not currently ask for the user's name.
- Added test


nickoneill (approved):
Thanks! this looks great, thanks for including the name bit even though we don't have a way to set it yet.

PR #492, opened 2025-09-26
Update circleci Xcode version to 16.4

Abizern:
Not pushing it to Xcode 26, but this matches the version in the readme.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
     # The `macos` block requests that builds will run on a machine running
     # macOS with the specified version of Xcode installed
     macos:
-      xcode: '13.4.0'
+      xcode: '16.4.0'
     steps:
     # Machine Setup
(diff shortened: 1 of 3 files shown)


nickoneill:
we don't use circle for testing anymore so we can probably get rid of that config. I'd like to move entirely to Xcode cloud but plugin validation was failing with the previous R.Swift version (half of my motivation to get rid of it). As I understand it we need some pre-build script config for Xcode cloud to disable plugin validation for it to work there, but it seemed easier to just do the work to replace it. It's failing with the same issue in Xcode cloud now: `Plugin “RswiftGenerateInternalResources” from package “Rswift” must be enabled before it can be used`

Abizern:
I saw those errors on Xcode Cloud. It says that it's fixed in the version that I upgraded it to, but it still isn't. I have already started on the Strings catalog work so that will be done soon.

PR #493, opened 2025-09-27
fix contact tests

nickoneill:
fixes the contact test that was looking for lowAccuracy in the reps response


Abizern (approved):
Tests pass locally, where there weren't before. Thank you.

PR #491, opened 2025-09-18
Remove R.swift in favor of native bundle strings

nickoneill:
Testing a semi-automated removal of R.swift that still maintains some safety of string-typed things

PR #494, opened 2025-10-09
Abizer/remove r swift

### LINKED ISSUE #488: Replace R.swift
I am frustrated with constant fake warnings in Xcode for R resources. Yeah sure, it always works when you build the whole thing but that doesn't fix the errors that pop up and go away without anything changing while you're editing.

Should we use the native string localization stuff in SwiftUI now? Something else?

Abizern:
Closes #488 
Alternative to  #491 

This PR removes the R.swift dependency for  #Strings

- Swapped out R.strings and conveniences to local definitions with comments.
- Pluralisation is handled in the Strings dictionary and not in code. e.g.
Screenshot 2025-10-09 at 22 18 08

- Colors and images accessed through Xcode generated convenience functions
- Readme updated to remove R.swift section
- Xcode project is set up to track strings. New strings will be added to the string catalog when defined and unused ones will be removed. This removes redundancy and duplication and gives visibility to all localised strains in the app. Strings catalogs allow jumping directly to the definition (from a context menu)
- SwiftUI views that take passed in strings converted to take StringResource instead, per current best practices.

SwiftUI is more about defining strings locally and making them available for localisation rather than defining the strings with convenience methods. Most strings are not reused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,23 +11,11 @@ This is the repository for the iOS app for [5Calls.org](https://5calls.org).
 
 ## Getting Started
 
-Install the dependencies:
+Clone the repository and resolve the Swift Packages. The gems installed by Bundler are more for distribution.
 
-```
-bundle install
-```
+### Strings
 
-## Using R.swift
-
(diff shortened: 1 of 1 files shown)


nickoneill:
thanks! sgtm

Issue #467, opened 2025-03-16
Pass dialed phone with result


nickoneill:
This will help us determine which phone numbers are unavailable if we want to push people dialing the DC number onto a local one for effectiveness.

We support recording which phone number was dialed with an outcome report which we used to support on iOS, but I think it was dropped in the redesign. Add the param `phone=4155551212` to the outcome POST.

I know we can't be 100% sure which phone was dialed but I think we can assume based on taps: keep tabs of the last phone number tapped and record that in the next outcome report.

raysj9:
Hi, is this issue still available?

nickoneill:
Yep @raysj9, feel free to pick it up!

PR #514, opened 2026-05-17
Pass dialed phone with result

raysj9:
Fixes Issue #467 

Passes the last clicked phone number with the report outcome POST operation. Also removes some unnecessary iOS 16 checks since the app no longer supports iOS 16. 

Verified that the dialed phone number is included in the request. 

Screenshot


cselin (approved):
✅ Looks good!

nickoneill:
Thank you @raysj9, sorry for the wait!

PR #516, opened 2026-09-06
Replace OneSignal with our own token registration API

nickoneill:
OneSignal is essentially cutting us off from using the free plan on October 1. We don't really use push regularly right now but to continue ensuring we have access to the data needed to potentially send pushes we need to migrate the new token sending to our own API and away from OneSignal. I wouldn't use OneSignal if we were starting this from scratch anyway since it's so trivial to manage this data and develop a UI for sending.

Tested this by confirming new hits on our API endpoint for token registration.


cselin (approved):
✅ Changes look great! It would be nice to increase the unit test coverage, especially since it comes almost for free assuming you are leveraging AI these days. Also, want to call out a testing flow where a user initially allows push notifications, but then turns them off to make sure the app handles it gracefully without automatically badgering them with additional permission requests.

